PR #20, opened 2026-03-23
Fix class-level key set mutation in AtomicData

atulcthakur:

## Description

  - `__node_keys__`,` __edge_keys__`, and `__system_keys__` on AtomicData were mutable sets shared across all instances. Calling `add_node_property()` on one instance permanently polluted the class-level set, affecting all existing and future instances. This PR aims to fix the class level pollution and keep `add_node_property() `at the instance level. 

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Changes Made

  - `nvalchemi/data/atomic_data.py` — Replaced mutable class-level sets with `ClassVar[frozenset] `defaults. Added `_init_key_sets` validator for per-instance copies.
  - `nvalchemi/data/batch.py` — `from_data_list()` reads key sets from instance instead of class.
  - `nvalchemi/data/datapipes/backends/zarr.py` — Updated class-level references to `_default_*_keys`.
  - `test/data/test_atomic_data.py` — Updated 3 tests for new names. Added 5 regression tests to verify existing and future instances are not affected with `add_node_property()`
  
Key change:

  Code that accesses `AtomicData.__node_keys__` (or `__edge_keys__`/`__system_keys__`) at the class level must use `AtomicData._default_node_keys` instead. Instance-level access (`data.__node_keys__`) is unchanged.

## Testing

## Checklist

## Additional Notes

> [!TIP]

### DIFF
--- a/test/data/test_atomic_data.py
+++ b/test/data/test_atomic_data.py
@@ -215,6 +215,18 @@ class TestAtomicDataPropertiesDict:
         c = _minimal_atomic_data(2)
         assert "custom_feat" not in c.__node_keys__
 
+    def test_add_edge_property_does_not_pollute_future_instances(self):
+        a = _minimal_atomic_data(3, num_edges=4)
+        a.add_edge_property("custom_edge", torch.randn(4, 2))
+        c = _minimal_atomic_data(3, num_edges=4)
+        assert "custom_edge" not in c.__edge_keys__
+
+    def test_add_system_property_does_not_pollute_future_instances(self):
+        a = _minimal_atomic_data(3)
+        a.add_system_property("custom_sys", torch.tensor([1.0]))
+        c = _minimal_atomic_data(3)
+        assert "custom_sys" not in c.__system_keys__
+
     def test_multiple_add_node_property_preserves_all_keys(self):
         data = _minimal_atomic_data(3)
         data.add_node_property("feat_a", torch.randn(3, 4))


nvalchemi/data/atomic_data.py
```diff
@@ -317,13 +318,29 @@ class AtomicData(BaseModel, DataMixin):
...
+        The class-level defaults are frozen to prevent accidental mutation.
+        Each instance gets its own mutable set so that ``add_node_property``
+        and friends only affect the instance they are called on.
+
+        This validator must run before ``check_node_consistency`` and
+        ``check_edge_consistency``, which iterate the key sets.
+        """
+        object.__setattr__(self, "__node_keys__", set(self._default_node_keys))
+        object.__setattr__(self, "__edge_keys__", set(self._default_edge_keys))
+        object.__setattr__(self, "__system_keys__", set(self._default_system_keys))
```
WardLT, line 348: This part confuses me a little. Why are we storing both the defaults and the "instance specific" keys in the same object?

(Could be an ignorant question. I'm still on-boarding with the data structures.)

test/data/test_atomic_data.py
```diff
@@ -181,6 +181,38 @@ def test_add_system_property(self):
...
+    def test_add_node_property_does_not_pollute_future_instances(self):
+        a = _minimal_atomic_data(3)
+        a.add_node_property("custom_feat", torch.randn(3, 4))
+        c = _minimal_atomic_data(2)
+        assert "custom_feat" not in c.__node_keys__
+
+    def test_class_level_defaults_are_immutable(self):
+        assert isinstance(AtomicData._default_node_keys, frozenset)
+        assert isinstance(AtomicData._default_edge_keys, frozenset)
+        assert isinstance(AtomicData._default_system_keys, frozenset)
```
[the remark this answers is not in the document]
laserkelvin: I agree with this as well - just for coverage's sake

test/data/test_atomic_data.py
```diff
@@ -123,19 +123,19 @@ def test_dtype(self):
 # Class-level keys
 # -----------------------------------------------------------------------------
 class TestAtomicDataKeys:
-    """Tests for __node_keys__, __edge_keys__, __system_keys__."""
+    """Tests for _default_node_keys, _default_edge_keys, _default_system_keys."""
 
     def test_node_keys_contains_positions(self):
-        assert "positions" in AtomicData.__node_keys__
-        assert "atomic_numbers" in AtomicData.__node_keys__
+        assert "positions" in AtomicData._default_node_keys
```
WardLT, line 136: Why are these tests for the `_defaults`? Are the defaults something that users should be accessing?

nvalchemi/data/atomic_data.py
laserkelvin: My interpretation of this is that the instance-specific values are being derived from the default ones. I don't know if it would make things clearer/cleaner, but probably doing `__node_keys__ | _default_node_keys` might look nicer for lookups

WardLT: I think I follow. Is "__node_keys__" instance specific and "_default_node_keys" globally-defined/not to be changed ever?

If so, I think I appreciate the idea of keeping those two conceptually separate

test/data/test_atomic_data.py
WardLT: (I think, based on Kelvin helping me above, I agree this is correct)

nvalchemi/data/atomic_data.py
atulcthakur: Thanks for the comments. 
The `_defaults` are class-level frozen attributes. We wouldn't want to update them when adding a `add_node_property`. The`object.__setattr__` basically copies the defaults into instance's `__node/edge/system_keys__`, so when we do `add_node_property`, we're really only mutating the instance and not polluting any other active instances. 

We're copying them into `__node/edge/system_keys__` because the code downstream uses them. But yes, I share the concern about code hygiene in the long-run.

laserkelvin: To @WardLT 's point, I think semantically it would make more sense then to rename the variables.

`__node/edge/system_keys__` uses dunder so it's *intended* to be class-level attribute, and so we should keep *them* as the default keys but add a new instance level attribute. We should then modify the downstream logic to base off the union of both sets.

WardLT: I'll second the idea of a new instance-specific attribute. Editing a `__`-dundered variable sets off alarm bells for me, fwiw

nvalchemi/data/atomic_data.py
```diff
@@ -317,13 +318,29 @@ class AtomicData(BaseModel, DataMixin):
...
+        return self
```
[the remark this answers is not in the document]
dallasfoster: I think this is probably what I would expect the behavior to be, right?

PR #27, opened 2026-03-25
refactor: standardize edge_index shape to (E, 2) across internal data path

zubatyuk:

## Description

Standardize `edge_index` to `(E, 2)` across the internal data path. Previously `edge_index` was `(2, E)` in some places and `(E, 2)` in others, requiring ad-hoc transposes scattered through `Batch.from_data_list`, `Batch.get_data`, Zarr write/read, and model adapters. This PR eliminates all internal transposes by making `(E, 2)` the single canonical shape for `AtomicData`, `Batch` (via `SegmentedLevelStorage` dim-0 slicing), and Zarr on-disk storage. Transposes are retained only at three external boundaries where the consuming code expects `(2, E)`: MACE model adapter, `nvalchemiops` neighbor list hook output, and `prepare_neighbors_for_model`.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Changes Made

- `nvalchemi/_typing.py`: `EdgeIndex` type alias `"2 E"` → `"E 2"`
- `nvalchemi/data/data.py`: `DataMixin.__cat_dim__` returns 0 for `edge_index`; `debug()` shape check expects `(E, 2)`; added comment that `__cat_dim__` is now largely vestigial
- `nvalchemi/data/atomic_data.py`: docstrings, Field description, `num_edges` → `shape[0]`, `check_edge_consistency` → `size(0)`
- `nvalchemi/data/batch.py`: `from_data_list` concatenates `edge_index` on dim 0 without transpose; `get_data` returns slices without transpose; clarified `_INDEX_KEYS` comment (offset-only, not shape-related)
- `nvalchemi/data/datapipes/backends/zarr.py`: `_get_cat_dim` returns 0 for `edge_index`; removed write/append transposes; added `_slice_edge_array` helper that enforces `cat_dim == 0` for all edge fields (raises `RuntimeError` for any field matching the vestigial `*index*`/`*face*` regex fallback)
- `nvalchemi/data/datapipes/dataset.py`: `edge_index.shape[0]` for `num_edges`
- `nvalchemi/models/mace.py`: retained `.T` at MACE boundary with updated comment
- `nvalchemi/dynamics/hooks/neighbor_list.py`: retained `.T.contiguous()` converting `nvalchemiops` `(2, E)` output to internal `(E, 2)`
- `nvalchemi/models/_ops/neighbor_filter.py`: added `.T.contiguous()` in COO path converting batch `(E, 2)` to `(2, E)` for `filter_neighbor_list`
- `nvalchemi/models/base.py`: updated `NeighborListFormat.COO` docstring to clarify internal vs model-boundary conventions
- `docs/userguide/data.md`, `examples/basic/01_data_structures.py`: `[2, n_edges]` → `[n_edges, 2]`
- All affected test files updated: `test_atomic_data`, `test_batch`, `test_data_mixin`, `test_zarr_datapipe`, `test_mace`, `test_neighbor_filter`, `test_sampler`
- New `test/dynamics/test_neighbor_list_hook.py`: 7 tests covering `(E, 2)` shape contract after hook execution (non-PBC, PBC, half-list, multi-graph)
- New `TestSliceEdgeArrayGuard` in `test/data/test_zarr_datapipe.py`: 4 tests verifying `_slice_edge_array` rejects `cat_dim != 0` keys and that `_load_sample`/`defragment` raise `RuntimeError` for a custom edge field named `face_index`

## Testing

## Checklist

## Additional Notes

- Zarr backward compatibility is not provided — existing Zarr files with `(2, E)` `edge_index` need re-export.
- `DataMixin.__cat_dim__` retains a vestigial regex fallback for `*index*`/`*face*` keys (returns -1). No current `AtomicData` attribute reaches it. The Zarr read paths (`_slice_edge_array`) now reject `cat_dim != 0` with a `RuntimeError`, so any future field matching the regex will fail loudly instead of silently producing a wrong slice.

### DIFF
--- a/nvalchemi/data/data.py
+++ b/nvalchemi/data/data.py
@@ -139,27 +139,6 @@ class DataMixin:
             if key in self:
                 yield key, self[key]
 
-    # NOTE: __cat_dim__ is now largely vestigial.  Batch uses
-    # SegmentedLevelStorage which always concatenates on dim 0, and the only
-    # remaining caller is DataMixin.num_edges.  The generic regex fallback
-    # for ``*index*`` / ``*face*`` keys is kept for backward compatibility
-    # but no active attribute in AtomicData reaches it.
-    def __cat_dim__(self, key: str, value: Any) -> int:
-        r"""Returns the dimension for which :obj:`value` of attribute
-        :obj:`key` will get concatenated when creating batches.
-
-        .. note::
-
-            This method is for internal use only, and should only be overridden
-            if the batch concatenation process is corrupted for a specific data
-            attribute.
-        """
-        if key == "edge_index":
-            return 0
-        if bool(re.search("(index|face)", key)):
-            return -1
-        return 0
-
     def __inc__(self, key: str, value: Any) -> int:
         r"""Returns the incremental count to cumulatively increase the value
         of the next attribute of :obj:`key` when creating batches.
@@ -187,18 +166,14 @@ class DataMixin:
         For undirected graphs, this will return the number of bi-directional
         edges, which is double the amount of unique edges.
         """
-        for key, item in {
-            "edge_index": self.edge_index,
-            "edge_attr": self.edge_attr,
-        }.items():
-            if item is not None:
-                return item.size(self.__cat_dim__(key, item))
-        for key, item in {
-            "adj": self.__dict__.get("adj"),
-            "adj_t": self.__dict__.get("adj_t"),
-        }.items():
-            if item is not None:
-                return item.nnz()
+        if self.edge_index is not None:
+            return self.edge_index.size(0)
+        if self.edge_attr is not None:
+            return self.edge_attr.size(0)
+        if (adj := self.__dict__.get("adj")) is not None:
+            return adj.nnz()
+        if (adj_t := self.__dict__.get("adj_t")) is not None:
+            return adj_t.nnz()
         return None
 
     def __apply__(self, item: Any, func: Callable[[Any], Any]) -> Any:
--- a/nvalchemi/data/datapipes/backends/zarr.py
+++ b/nvalchemi/data/datapipes/backends/zarr.py
@@ -210,13 +210,11 @@ def _get_field_level(key: str) -> str:
             return "atom"
 
 
-# NOTE: the generic *index*/*face* regex fallback returning -1 is vestigial.
-# No current AtomicData attribute reaches it, and the Zarr read paths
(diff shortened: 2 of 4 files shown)


nvalchemi/data/datapipes/backends/zarr.py
```diff
@@ -747,7 +743,6 @@ def defragment(self) -> None:
                 elif level == "edge":
                     cat_dim = _get_cat_dim(key)
                     if cat_dim == -1:
-                        # edge_index: shape [2, E]
                         new_core_data[key].append(arr[:, edge_start:edge_end])
                     else:
                         new_core_data[key].append(arr[edge_start:edge_end])
```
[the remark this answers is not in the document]
zubatyuk: Fixed in fa89e30a08026a1f9f1c4b88533625697cc1ebc9. Updated PR message.

nvalchemi/models/_ops/neighbor_filter.py
```diff
@@ -407,7 +407,8 @@ def prepare_neighbors_for_model(
 
     # Sub-case B: have list — filter if needed.
     if has_list:
-        nl = data.edge_index
+        # Batch stores (E, 2); filter_neighbor_list operates on (2, E)
+        nl = data.edge_index.T.contiguous()
```
atulcthakur, line 411: Do you really need `.contiguous()` here? It can cause reallocation of the entire tensor if not already contiguous in memory. Usually it's a performance chokepoint.

zubatyuk: I would leave it like this for safety. Later we can do a pass to resolve such inefficiencies.

nvalchemi/data/data.py
```diff
@@ -139,6 +139,11 @@ def __call__(self, *keys: str) -> Iterator[tuple[str, Any]]:
             if key in self:
                 yield key, self[key]
 
+    # NOTE: __cat_dim__ is now largely vestigial.  Batch uses
```
laserkelvin: From this comment, it sounds like we can potentially deprecate this method, and have `DataMixin.num_edges` possibly determine edge counts differently. I would like to not have to maintain unused methods so let's see if we can remove it

nvalchemi/models/_ops/neighbor_filter.py
laserkelvin: @atulcthakur it won't be contiguous after transpose, and IIRC it can affect `warp` kernel usage as the stride changes so I agree with @zubatyuk that we should keep it here for now

PR #55, opened 2026-04-08
Composable autograd pipeline and model wrappers

zubatyuk:

## Description

Replace the additive-only model composition layer with a composable calculator pipeline that supports external autograd, dependent model wiring, and automatic neighbor-list planning.

## Type of Change

- [x] Performance improvement
- [ ] Documentation update

## Changes Made

- Composable model pipeline with external autograd: `ComposableModelWrapper` now orchestrates an ordered node list through `PipelineContext`, with one explicit autograd region for energy-to-forces/stresses differentiation via `DerivativeStep`.
- `DerivativeStep` supports forces, stresses, and arbitrary Jacobian products (e.g. dipoles → BEC tensors) in a single composable step.
- `ModelConfig` replaces `ModelCard` + mutable `ModelConfig` with a single frozen dataclass declaring the full execution contract (required/optional inputs, outputs, additive outputs, autograd participation, neighbor requirements).
- `BaseModelMixin` simplified from ABC with abstract methods to a concrete mixin with `spec: ModelConfig`, `adapt_input(batch, ctx)`, `adapt_output(raw, compute=)`.
- Automatic neighbor-list synthesis: `NeighborListBuilder` nodes are inserted before their first consumer, one per distinct (cutoff, half_list) pair, with density-adaptive capacity management.
- `NeighborList` payload with `.adapt()` for on-the-fly format conversion (matrix ↔ COO) and cutoff narrowing.
- `wire_output(source, target, mapping)` for explicit dependent composition (e.g. charge predictor → Ewald).
- New model: `AIMNet2Wrapper` wrapping AIMNetCentral with deferred optional dependency import.
- New model: `DSFModelWrapper` for damped shifted force Coulomb electrostatics (periodic and non-periodic).
- `MACEWrapper` now accepts `str | Path | nn.Module` directly, with built-in dtype conversion, cuEquivariance conversion, and `torch.compile` support.
- All model wrappers refactored to Pydantic config classes (`LennardJonesConfig`, `EwaldCoulombConfig`, `PMEConfig`, `DFTD3Config`, `DSFCoulombConfig`) with `_resolve_config` pattern for flexible construction.
- `make_neighbor_hooks()` removed from models — decouples models from dynamics.
- Removed `registry.py` (model registry).

## Testing

## Checklist

## Additional Notes

Breaking changes:
- `ModelCard` removed; use `model.spec` (a `ModelConfig` frozen dataclass) instead of `model.model_card`.
- Mutable `ModelConfig` with `compute_forces`/`compute_stresses` booleans removed; pass `compute={"energies", "forces"}` to `forward()` instead.
- `model.make_neighbor_hooks()` removed; neighbor lists are now managed by the composable pipeline's `NeighborListBuilder` nodes.
- `ComposableModelWrapper.forward()` signature changed from `(data: Batch, **kwargs) -> OrderedDict` to `(batch: Batch, *, compute=None, detach=True) -> dict`.
- Individual wrapper `forward()` signatures changed from `(data: Batch, **kwargs) -> ModelOutputs` to `(data: dict[str, Tensor]) -> dict[str, Tensor]`.
- `embedding_shapes`, `compute_embeddings`, `export_model` abstract methods removed from `BaseModelMixin`.


nvalchemi/models/derivatives.py
```diff
@@ -0,0 +1,401 @@
...
+            enabled_builtins.add("stresses")
+        if specs is not None:
+            overlap = enabled_builtins & set(specs)
+            if overlap:
+                raise ValueError(
+                    f"Specs keys overlap with builtin outputs: {sorted(overlap)}"
+                )
+            for output_name, spec_value in specs.items():
+                normalized = self._normalize_output_spec(output_name, spec_value)
+                derivatives[output_name] = DerivativeSpec(*normalized)
```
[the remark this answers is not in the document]
zubatyuk: Fixed in 31b8499186a1392711446bcb40c82e76f84b2b69.

nvalchemi/models/ewald.py
```diff
@@ -14,449 +14,341 @@
...
+    additive_outputs=frozenset({"energies", "forces", "stresses"}),
+    use_autograd=True,
+    pbc_mode="pbc",
+    neighbor_config=NeighborConfig(
+        source="external",
+        cutoff=15.0,
+        format="matrix",
+        half_list=False,
+    ),
+)
```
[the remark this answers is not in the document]
zubatyuk: Fixed in 31b8499186a1392711446bcb40c82e76f84b2b69.

nvalchemi/models/pme.py
```diff
@@ -15,486 +15,399 @@
...
+            "num_neighbors",
+        }
+    ),
+    optional_inputs=frozenset({"neighbor_shifts"}),
+    outputs=frozenset({"energies", "forces"}),
+    optional_outputs={"stresses": frozenset({"cell", "pbc"})},
+    additive_outputs=frozenset({"energies", "forces", "stresses"}),
+    use_autograd=True,
+    pbc_mode="pbc",
+    neighbor_config=NeighborConfig(
```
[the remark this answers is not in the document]
zubatyuk: Fixed in 31b8499186a1392711446bcb40c82e76f84b2b69.

nvalchemi/models/dsf.py
```diff
@@ -0,0 +1,253 @@
...
+    additive_outputs=frozenset({"energies", "forces", "stresses"}),
+    use_autograd=True,
+    pbc_mode="any",
+    neighbor_config=NeighborConfig(
+        source="external",
+        cutoff=15.0,
+        format="coo",
+        half_list=False,
+    ),
+)
```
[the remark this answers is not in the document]
zubatyuk: Fixed in 31b8499186a1392711446bcb40c82e76f84b2b69.

nvalchemi/models/aimnet2.py
```diff
@@ -0,0 +1,310 @@
...
+    outputs=frozenset({"energies", "node_charges"}),
+    additive_outputs=frozenset({"energies"}),
+    use_autograd=True,
+    autograd_outputs=frozenset({"node_charges"}),
+    pbc_mode="any",
+)
+
+_AIMNET_OPTIONAL_INPUTS_WITH_SPIN = frozenset(
+    {"cell", "pbc", "graph_charges", "graph_spins"}
+)
```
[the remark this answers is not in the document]
zubatyuk: Fixed in 31b8499186a1392711446bcb40c82e76f84b2b69.

nvalchemi/models/dftd3.py
```diff
@@ -15,406 +15,343 @@
...
+    "wb97m": {"a1": 0.5660, "a2": 3.1280, "s6": 1.0, "s8": 0.3908},
+    "wb97x": {"a1": 0.0000, "a2": 5.4959, "s6": 1.0, "s8": 0.2641},
+    "r2scan": {"a1": 0.49484001, "a2": 5.73083694, "s6": 1.0, "s8": 0.78981345},
+}
+DFTD3_ARCHIVE_URL = (
+    "https://github.com/dftd3/simple-dftd3/archive/refs/heads/main.tar.gz"
+)
+DFTD3_ARCHIVE_SHA256 = (
+    "0eb3e36bfb24dcd9bb1d1bece1531216b59539a8fde17ee80224af0653c92aa3"
 )
```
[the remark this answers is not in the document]
zubatyuk: Fixed in 31b8499186a1392711446bcb40c82e76f84b2b69.

nvalchemi/models/dftd3.py
```diff
@@ -15,406 +15,343 @@
...
     return {
-        "rcov": torch.from_numpy(rcov),
-        "r4r2": torch.from_numpy(r4r2),
-        "c6ab": torch.from_numpy(c6ab),
-        "cn_ref": torch.from_numpy(cn_ref),
+        "rcov": rcov,
+        "r4r2": r4r2,
+        "c6ab": c6_raw.reshape(95, 95, 5, 5),
+        "cn_ref": cn_raw.reshape(95, 95, 5, 5),
     }
```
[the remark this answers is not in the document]
zubatyuk: Fixed in 31b8499186a1392711446bcb40c82e76f84b2b69.

nvalchemi/models/aimnet2.py
```diff
@@ -0,0 +1,310 @@
...
+
+AIMNet2ModelConfig = ModelConfig(
+    required_inputs=frozenset({"positions", "atomic_numbers"}),
+    optional_inputs=frozenset({"cell", "pbc", "graph_charges"}),
+    outputs=frozenset({"energies", "node_charges"}),
+    additive_outputs=frozenset({"energies"}),
+    use_autograd=True,
+    autograd_outputs=frozenset({"node_charges"}),
+    pbc_mode="any",
+)
```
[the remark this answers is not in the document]
zubatyuk: Fixed in 31b8499186a1392711446bcb40c82e76f84b2b69.

nvalchemi/models/base.py
```diff
@@ -12,601 +12,642 @@
...
-warnings.simplefilter("once", UserWarning)
+    def __bool__(self) -> bool:
+        """Treat the sentinel as falsy."""
 
+        return False
 
-class NeighborListFormat(str, Enum):
-    """Storage format for neighbor data written to the batch.
+
+_UNSET: Any = _UnsetType()
```
dallasfoster, line 67: This sentinel pattern adds complexity that benefits only the config resolution path (`_resolve_config`). Standard Python uses `None` as a sentinel, or a simple `_UNSET = object()`. The full class with `__new__`, `__repr__`, `__bool__` is overengineered for what amounts to a missing-value marker.

nvalchemi/models/base.py
```diff
@@ -12,601 +12,642 @@
...
-    half_list : bool
-        If ``True``, each pair ``(i, j)`` with ``i < j`` appears only once.
-        Newton's third law is applied inside the interaction kernel to recover
-        forces on both atoms.  Defaults to ``False``.
-    skin : float
-        Verlet skin distance.  The neighbor list is only rebuilt when any atom
-        has moved more than ``skin / 2`` since the last build.  Set to ``0.0``
-        (default) to rebuild every step.
+    source : {"none", "internal", "external"}
+        Declares who is responsible for neighbor-list construction.
```
dallasfoster, line 97: I do not think this is necessary. What is the functional difference between "none" and "internal"?

nvalchemi/models/base.py
```diff
@@ -12,601 +12,642 @@
...
-        ),
-    ]
-
-
-class ModelCard(BaseModel):
+    required_inputs: frozenset[str] = frozenset()
+    optional_inputs: frozenset[str] = frozenset()
+    outputs: frozenset[str] = frozenset()
+    optional_outputs: dict[str, frozenset[str]] = field(default_factory=dict)
+    additive_outputs: frozenset[str] = frozenset()
```
dallasfoster, line 173: I feel like again, these two add unnecessary complexity to our code.

nvalchemi/models/base.py
```diff
@@ -12,601 +12,642 @@
...
+    use_autograd: bool = True
+    autograd_inputs: frozenset[str] = frozenset()
+    autograd_outputs: frozenset[str] = frozenset()
+    pbc_mode: Literal["non-pbc", "pbc", "any"] | None = None
```
dallasfoster, line 177: I'm not sure I like this better than the current formulation and I don't know why you felt the need to change it.

nvalchemi/models/base.py
```diff
@@ -12,601 +12,642 @@
...
-    Model card for a given model.
 
-    This model card is a Pydantic model that contains information about the model's
-    capabilities and requirements.
+    filtered = {key: value for key, value in overrides.items() if value is not _UNSET}
+    if config is not None and filtered:
+        return config.model_copy(update=filtered)
+    if config is not None:
+        return config
+    return config_cls(**filtered)
```
dallasfoster, line 253: Again, this feels like unnecessary complexity and maintenance burden.

nvalchemi/models/base.py
```diff
@@ -12,601 +12,642 @@
...
+        scaling = scaling.requires_grad_(True)
+
+        if scaling.ndim == 3:
+            atom_scaling = torch.index_select(scaling, 0, batch_idx)
+            positions_scaled = (positions.unsqueeze(1) @ atom_scaling).squeeze(1)
+        else:
+            positions_scaled = positions @ scaling
+
+        cell_scaled = cell @ scaling if cell.ndim == 2 else torch.bmm(cell, scaling)
+        return positions_scaled, cell_scaled, scaling
```
dallasfoster, line 368: This is a significant amount of machinery (~90 lines) to manage what is essentially two dicts and a boolean flag. The session encapsulates the stress scaling setup (`_prepare_stress_scaling`), which is useful, but wrapping it in a class with 8 methods and 4 properties makes it harder to understand than a few standalone functions.

I would rather take the approach to provide `prepare_strain()` as a standalone function that any caller (wrapper, pipeline, user script) can use. The pipeline handles the session-like state internally without a dedicated class. This keeps the concept count low.

The core question: does the `_AutogradSession` abstraction help a user who is debugging a stress computation? Or does it add a layer they have to dig through? Does it really make our lives easier?

nvalchemi/models/base.py
```diff
@@ -12,601 +12,642 @@
...
+    def get_autograd_target(self, name: str) -> Tensor:
+        """Return one tracked derivative target."""
 
-        Parameters
-        ----------
-        model_output : Any
-            Raw output from the external model
-        data : AtomicData | Batch
-            Original input data (may be needed for context/metadata)
+        return self._autograd.get_target(name)
```
dallasfoster, line 509: `PipelineContext` is the central state container for the composable runtime. While it's well-structured, it introduces a concept that every model wrapper author must understand — `ctx.resolve(key, batch)` instead of `getattr(batch, key)`. This changes the fundamental mental model from "models receive a batch" to "models receive a batch AND a context."

In our approach, models always receive a `Batch` and we handle wiring by writing upstream outputs onto the batch (via `object.__setattr__`) before calling downstream models. This is admittedly less clean, but it means model authors never need to know about a pipeline context. Their `forward()` looks the same whether the model is standalone or composed.

The tradeoff is simplicity of the wrapper API vs. cleanliness of the composition internals. We've chosen the former.

Think that every additional classes make it harder to add features and functionalities as a user and as a developer in the code base later.

nvalchemi/models/base.py
```diff
@@ -12,601 +12,642 @@
...
-    def export_model(self, path: Path, as_state_dict: bool = False) -> None:
-        """
-        Export the current model without the ``BaseModelMixin`` interface.
-
-        The idea behind this method is to allow users to use the trained
-        model with the same interface as the corresponding 'upstream' version,
-        so that they can re-use validation code that might have been written
-        for the upstream case (e.g. ``ase.Calculator`` instances).
+        dict[str, Any]
+            Input mapping passed to ``forward``.
```
dallasfoster, line 584: This is where the `PipelineContext` dependency surfaces for users. Every wrapper's `adapt_input` must now accept a `ctx` parameter and use `ctx.resolve()` instead of `getattr(batch, ...)`. This means:

1. A standalone model call requires constructing a `PipelineContext` even when there's no pipeline.
2. Every existing wrapper must be updated to use the context resolution pattern.
3. Users writing custom wrappers must learn the context API.

Our approach keeps `adapt_input(data, **kwargs)` — the model doesn't know or care whether it's inside a pipeline.

nvalchemi/models/composable.py
dallasfoster, line 1: This file is ~1500 lines (composable.py alone), introducing 8+ new classes/types: `_ComposableModel`, `_ComposableContract`, `_WiringEdge`, `_NodePlan`, `_CompiledPlan`, `_ExecutionRequest`, `_ComposableRuntimeModel`, and `ComposableModelWrapper`. Each of these is well-designed individually, but the aggregate concept count is high.

Greptile already flagged multiple P1 bugs in this file (missing `"batch"` key in several model configs). The complexity is contributing to these kinds of mistakes.

nvalchemi/models/composable.py
```diff
@@ -14,305 +14,1943 @@
...
+_INTERNAL_NEIGHBOR_KEYS = frozenset(
+    {
+        "edge_index",
+        "neighbor_ptr",
+        "unit_shifts",
+        "neighbor_matrix",
+        "num_neighbors",
+        "neighbor_shifts",
+    }
+)
```
dallasfoster, line 84: I don't love having these default keys listed like this.

nvalchemi/models/composable.py
```diff
@@ -14,305 +14,1943 @@
...
 
-    Parameters
+
+@runtime_checkable
+class _ComposableModel(Protocol):
+    """Typing protocol for one composable executable model."""
+
+    spec: ModelConfig
+
+    def forward(self, data: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]: ...
```
dallasfoster, line 98: This seems unnecessary to me.

nvalchemi/models/composable.py
```diff
@@ -14,305 +14,1943 @@
...
+                else None
+            )
+            self.output_publish_plan[key] = (
+                mapped_key,
+                is_additive,
+                key in spec.autograd_outputs,
+                qualified_key,
+            )
+        if self.validate_runtime_outputs:
+            self.runtime_validator = _make_runtime_validator(spec)
```
dallasfoster, line 259: This is an over complicated and over expansive optimization for this repo at this point.

nvalchemi/models/composable.py
```diff
@@ -14,305 +14,1943 @@
...
+    produced : frozenset[str]
+        Output keys always produced (e.g. ``{"forces", "stresses"}``).
+    conditional : dict[str, frozenset[str]]
+        Outputs that are only produced when certain batch keys are present
+        (e.g. ``{"stresses": {"cell", "pbc"}}``).
+    """
+
+    step: DerivativeStep
+    produced: frozenset[str]
+    conditional: dict[str, frozenset[str]]
```
dallasfoster, line 279: Again, I think an over complication that is unnecessary.

nvalchemi/models/derivatives.py
```diff
@@ -0,0 +1,461 @@
...
+            raise ValueError(
+                "DerivativeSpec.mode must be one of "
+                f"{sorted({'grad', 'jacobian', 'stress'})!r}, got {self.mode!r}"
+            )
+
+    @property
+    def grad_mode(self) -> Literal["grad", "jacobian"]:
+        """Return the grouped autograd mode used internally."""
+
+        return "grad" if self.mode in {"grad", "stress"} else "jacobian"
```
dallasfoster, line 88: The common theme here is over engineering. Your code is throwing in too many additional classes, too many additional features without pausing to understand the long-term impact.

nvalchemi/models/derivatives.py
```diff
@@ -0,0 +1,461 @@
...
+    stresses
+        Request stresses from ``energies`` with respect to the internal
+        ``cell_scaling`` target.
+    specs
+        Optional custom derivative products keyed by output name.
+    jacobian_chunk_size
+        Optional chunk size used when materializing large Jacobians.
+    create_graph
+        Whether to retain higher-order gradient information.
+    """
```
dallasfoster, line 127: For how much machinery you are building this class and method are surprisingly inflexible and difficult to reason about.

nvalchemi/models/derivatives.py
dallasfoster, line 1: This is the most important file to discuss because it represents a fundamental design choice: should derivative computation be a framework-owned abstraction?

`DerivativeStep` is powerful — it handles forces, stresses, arbitrary Jacobians, chunked Hessians, grouped autograd execution, and `retain_graph` management. But this power comes at a cost:

1. Users lose control. A user who wants forces computed with `create_graph=True` for a training loss, or who wants to chunk their Hessian differently, or who wants to use `torch.autograd.functional.jacobian` instead of manual looping, must work around the `DerivativeStep` API rather than just writing `torch.autograd.grad`.

2. We own the maintenance. Every `torch.autograd` edge case, every interaction with `torch.compile`, every new derivative product — we must handle it in this class. `torch.autograd.grad` is well-maintained by the PyTorch team; we don't need to wrap it.

3. The abstraction is leaky. `DerivativeSpec` must expose `negate`, `accumulate`, and `grad_mode` because the generic framework can't handle the semantic differences between forces (-dE/dr), stresses (dE/d_strain / V), and Jacobians. Each derivative product has its own post-processing — which is exactly what a user would write in their own `forward()`.

nvalchemi/models/neighbors.py
dallasfoster, line 1: My biggest complaint about this file (beyond the fact that I still think you are over engineering and trying to put too many features into one PR) is that this means that the neighbor list infrastructure exists in two places (hooks and model composition). Because of future aspects of our roadmap, I am unwilling to make too many changes to neighbor list handling right now.

dallasfoster (commented):
There are a lot of good ideas here, in particular we should consider adopting an AdaptiveNeighborList in a *future* PR, but it is surely out of scope right here. The vast majority of this code, however, adds too much complexity right now.

1. `PipelineContext` — changes the wrapper contract from `forward(batch)` to `forward(batch, ctx)`
2. `DerivativeStep` / `DerivativeSpec` — framework-owned derivative computation that takes control away from users
3. `_NodePlan` execution planning — premature optimization for 2-3 model pipelines
4. Internal neighbor list management — duplicates dynamics hook infrastructure
5. `additive_outputs` / `optional_outputs` on ModelConfig — conflates composition concepts with model capabilities

The core philosophical difference: this PR optimizes for the power of the composition framework, and while we want some degree of power we also want to optimize for the flexibility and simplicity of the individual wrapper author's experience and the division of responsibilities between us and the users. Both are valid — but for a package whose primary users are ML researchers wrapping new MLIPs, we believe the wrapper author's experience should take priority.

PR #99, opened 2026-05-29
make PipelineModelWrapper trainable

ys-teh:

## Description

Update PipelineModelWrapper so it can be used in training mode.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Changes Made

- Update PipelineModelWrapper when self.training=True and grad is enabled.
- Add MockTrainableAutogradEnergyModel and test_training_preserves_force_graph_for_backward for testing.

## Testing

- [ ] New tests added for new functionality meets coverage expectations?

## Checklist

- [ ] I have performed a self-review of my code

## Additional Notes

> [!TIP]

### DIFF
--- a/nvalchemi/models/pipeline.py
+++ b/nvalchemi/models/pipeline.py
@@ -682,6 +682,7 @@ class PipelineModelWrapper(nn.Module, BaseModelMixin):
         # Determine what derivatives are requested beyond energies.
         if isinstance(data, AtomicData):
             data = Batch.from_data_list([data])
+        training_with_grad = self.training and torch.is_grad_enabled()
         requested_derivatives = self.model_config.active_outputs - {"energy"}
 
         # Collect all autograd_inputs that need requires_grad
@@ -703,6 +704,7 @@ class PipelineModelWrapper(nn.Module, BaseModelMixin):
         autograd_groups = [g for g in self.groups if g.use_autograd]
         group_outputs: list[ModelOutputs] = []
         autograd_count = len(autograd_groups)
+        effective_autograd_count = autograd_count + int(training_with_grad)
         autograd_idx = 0
 
         for group in self.groups:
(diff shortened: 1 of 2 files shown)


test/models/test_pipeline.py
laserkelvin, line 1: Can you add one or two cases to the unit test cases that will include stress, and/or some scenario where you might need to use different ("non-standard"?) autograd input/outputs to train?

nvalchemi/models/pipeline.py
laserkelvin, line 939: I want to say the potential issue with this is that the `positions` gradients will be tracked even if we only want to train on energies. Ideally we can fix in this PR, but if it gets too complicated we can open an issue to track.

At the very least, you should make sure there are no memory leaks, but the best case scenario is to not have the overhead if we can get around it.

nvalchemi/models/pipeline.py
```diff
@@ -934,6 +943,7 @@ def _build_autograd_group_output(
                         displacement=strain.displacement,
                         orig_cell=strain.cell_for_stress,
                         retain_graph=retain_graph,
```
laserkelvin, line 945: Do the `retain_graph` flags need to factor in `training` as well?

laserkelvin (commented):
Some comments for your consideration - functionally I think I would be able to approve, but we should get these things addressed if possible.

nvalchemi/models/pipeline.py
ys-teh: I think it's already account for since "effective_retain = retain_graph or training" is set in autograd_forces and autograd_stresses.
https://github.com/NVIDIA/nvalchemi-toolkit/blob/df001caa30a8494d79b3c5348f5c7fd151b51fbb/nvalchemi/models/_utils.py#L76

test/models/test_pipeline.py
ys-teh: Added test including stress.

laserkelvin (approved):

laserkelvin:
/ok to test [e0015d3](https://github.com/NVIDIA/nvalchemi-toolkit/pull/99/commits/e0015d3735ae378a2b587ec4dca6a0ab3e55c875)

PR #110, opened 2026-06-09
Resolve ewald cache shape issue

ys-teh:

## Description

The Ewald/PME wrappers used torch.allclose(cell, self._cached_cell) to detect cell changes for cache invalidation. This assumes the current and cached cell tensors have identical shape, dtype, and device. When the same wrapper instance sees a different batch size, for example between training and validation batches, cell.shape changes from (B1, 3, 3) to (B2, 3, 3). In that case torch.allclose raises instead of returning False, so the cache cannot be safely invalidated and model training is affected.

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [ ] Performance improvement
- [ ] Documentation update

## Changes Made

- Implements a core cache invalidation fix for Ewald and PME cell caches.
	- Added _cell_cache_needs_update() to both EwaldModelWrapper and PMEModelWrapper.
	- The helper now treats missing cache, shape mismatch, device mismatch, dtype mismatch, or changed cell values as stale cache conditions.
	- Updated the forward paths to call this helper before recomputing Ewald/PME cache state.
	- Adds regression tests for both wrappers covering: missing cached cell, identical cached cell reuse, train/validation batch-size shape mismatch, same-shape changed cell values, dtype mismatch

## Testing

- [X] Unit tests pass locally (`make pytest`)
- [X] Linting passes (`make lint`)
- [ ] New tests added for new functionality meets coverage expectations?

## Checklist

- [X] I have read and understand the [Contributing Guidelines](../CONTRIBUTING.md)
- [X] I have updated the [CHANGELOG.md](../CHANGELOG.md)
- [X] I have performed a self-review of my code
- [X] I have added docstrings to new functions/classes

## Additional Notes

> [!TIP]

### DIFF
--- a/nvalchemi/models/_utils.py
+++ b/nvalchemi/models/_utils.py
@@ -49,6 +49,8 @@ __all__ = [
 def cell_cache_needs_update(
     cell: LatticeVectors,
     cached_cell: LatticeVectors | None,
+    rtol: float = 1e-5,
+    atol: float | None = None,
 ) -> bool:
     """Return ``True`` when ``cell`` is incompatible with ``cached_cell``.
 
@@ -59,19 +61,32 @@ def cell_cache_needs_update(
     cached_cell : torch.Tensor | None
         Previously cached cell tensor, or ``None`` when no cell has been
         cached yet.
+    rtol : float, optional
+        Relative tolerance passed to :func:`torch.allclose`.
+        Defaults to ``1e-5``.
+    atol : float or None, optional
+        Absolute tolerance passed to :func:`torch.allclose`.
+        When ``None`` (default), uses
+        ``max(1e-6, torch.finfo(cell.dtype).eps)``.
 
     Returns
     -------
     bool
         ``True`` if the cache should be refreshed.
     """
-    return (
-        cached_cell is None
-        or cell.shape != cached_cell.shape
-        or cell.device != cached_cell.device
-        or cell.dtype != cached_cell.dtype
-        or not torch.allclose(cell, cached_cell, rtol=1e-6, atol=1e-9)
-    )
+    if atol is None:
+        atol = max(1e-6, torch.finfo(cell.dtype).eps)
+
+    if cached_cell is None or cell.shape != cached_cell.shape:
+        return True
+    # Check device and dtype for compatibility with torch.allclose
+    if cell.device != cached_cell.device:
+        return True
+    if cell.dtype != cached_cell.dtype:
+        return True
+    if not torch.allclose(cell, cached_cell, rtol=rtol, atol=atol):
+        return True
+    return False
 
 
 def autograd_forces(
--- a/nvalchemi/models/_utils.py
+++ b/nvalchemi/models/_utils.py
@@ -40,9 +40,38 @@ __all__ = [
     "autograd_forces",
     "autograd_forces_and_stresses",
     "autograd_stresses",
+    "cell_cache_needs_update",
     "prepare_strain",
     "sum_outputs",
 ]
+
+
+def cell_cache_needs_update(
+    cell: torch.Tensor,
+    cached_cell: torch.Tensor | None,
+) -> bool:
+    """Return ``True`` when ``cell`` is incompatible with ``cached_cell``.
+
+    Parameters
+    ----------
+    cell : torch.Tensor
+        Current cell tensor.
+    cached_cell : torch.Tensor | None
+        Previously cached cell tensor, or ``None`` when no cell has been
+        cached yet.
+
+    Returns
+    -------
+    bool
+        ``True`` if the cache should be refreshed.
+    """
+    return (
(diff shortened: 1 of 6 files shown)


nvalchemi/models/ewald.py
```diff
@@ -210,6 +210,17 @@ def invalidate_cache(self) -> None:
...
+    def _cell_cache_needs_update(self, cell: torch.Tensor) -> bool:
+        """Return ``True`` when *cell* is incompatible with the cached cell."""
+        cached_cell = self._cached_cell
+        return (
+            cached_cell is None
+            or cell.shape != cached_cell.shape
+            or cell.device != cached_cell.device
+            or cell.dtype != cached_cell.dtype
+            or not torch.allclose(cell, cached_cell, rtol=1e-6, atol=1e-9)
+        )
```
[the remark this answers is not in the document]
ys-teh: Resolved this by creating a function called `call_cache_needs_update` in `_utils.py` so this function can be used by both `EwaldModelWrapper` and `PMEModelWrapper`.

nvalchemi/models/_utils.py
```diff
@@ -40,11 +40,40 @@
...
     "autograd_forces_and_stresses",
     "autograd_stresses",
+    "cell_cache_needs_update",
     "prepare_strain",
     "sum_outputs",
 ]
 
 
+def cell_cache_needs_update(
+    cell: torch.Tensor,
```
laserkelvin: Do you mind adding the appropriate `jaxtyping` shape annotations?

You might need to have them as separate hints to denote that the cached and the incoming cells can be different shapes

nvalchemi/models/_utils.py
```diff
@@ -40,11 +40,40 @@
...
+
+    Returns
+    -------
+    bool
+        ``True`` if the cache should be refreshed.
+    """
+    return (
+        cached_cell is None
+        or cell.shape != cached_cell.shape
+        or cell.device != cached_cell.device
```
laserkelvin: Do they need to be rebuilt if the device and dtype are different? Are there circumstances where this would be done by intention?

I just don't want to spend time recomputing if the cell is actually the same, but just on a different device (dtype might be acceptable)

nvalchemi/models/_utils.py
```diff
@@ -40,11 +40,40 @@
...
+        or cell.dtype != cached_cell.dtype
+        or not torch.allclose(cell, cached_cell, rtol=1e-6, atol=1e-9)
```
laserkelvin: The tolerances might depend on dtype. I think these might be too tight

laserkelvin (changes requested):
In general looks good to me, just have some minor things to discuss

nvalchemi/models/_utils.py
ys-teh: The tolerances were copied from here: https://github.com/NVIDIA/nvalchemi-toolkit/blob/main/nvalchemi/models/pme.py#L422
Is there a good value to put here? How about something like below?
rtol=1e-5
atol = max(1e-6, machine epsilon based on dtype)

laserkelvin: Yeah I think that's probably a lot better

IMO ideally we could expose this setting on the user side, so perhaps you could add them as kwargs to the function and that way users could specify how aggressive they want to be with re-computation.

nvalchemi/models/_utils.py
ys-teh: We don't need to rebuild if the device and dtype are different. But torch.allclose requires the same device and dtype. So I was taking to a shortcut / playing safe by adding both device and dtype to the check. Maybe we can have separate checks for device and dtype and then update the cache device and dtype if they are not consistent instead of asking for rebuilding straightaway?

nvalchemi/models/_utils.py
ys-teh: Updated, thanks for the reminder.

nvalchemi/models/_utils.py
laserkelvin: We can leave it for now - I think from your intention of just proactively guarding in order to call `torch.allclose` is justified

I might suggest breaking it up into multiple statements for the sake of reading and debugging though:

```python
# if you were to do an interactive debugger you would know which case it
# resolves so you can fix it
if cell.shape != cached_cell.shape:
    return True
if cell.dtype != cached_cell.dtype:
    return True
return torch.allclose(...)
```

ys-teh:
Thanks, I addressed all the comments including allowing users to change the tolerances when initializing the wrappers. All related tests passed.

laserkelvin (approved):
LGTM

laserkelvin:
/ok to test [b0af7b4](https://github.com/NVIDIA/nvalchemi-toolkit/pull/110/commits/b0af7b4c2c72ef359dbac6505bd1983006d016bf)

PR #113, opened 2026-06-15
fix: preserve system IDs during inflight refill

### LINKED ISSUE #112: 🐛[BUG]: BaseDynamics.refill_check overwrites replacement system_id during inflight refill
### Version

0.1.0 commit 176d4fe8c099566ff0363000c80e3b2b591ba7ce

Source

### Describe the issue

`SizeAwareSampler` stamps a monotonically increasing `system_id` onto each `AtomicData` sample it returns. During inflight batching, `BaseDynamics.refill_check()` removes graduated systems, appends replacement systems from the sampler, and then rebuilds registered bookkeeping tensors.

The sampler appears to assign the IDs correctly. For example, in a three-system dataset, the initial active batch gets IDs `[0, 1]`, and the next replacement `AtomicData` gets ID `[2]`. However, The `system_id` is lost after replacement systems are appended in the fused_stage pipeline.

`refill_check()` rebuilds bookkeeping tensors from default values. It copies IDs for systems that stayed in the active batch, but it does not copy the IDs that were already present on newly appended replacement systems. As a result, replacement `system_id` values are overwritten with the default `-1`.

Expected behavior: replacement systems should preserve the `system_id` assigned by `SizeAwareSampler`.

Actual behavior: replacement systems' `system_id` values are overwritten with `-1`.

```python
import torch
from nvalchemi.data import AtomicData
from nvalchemi.dynamics import FusedStage, NVTLangevin, SizeAwareSampler


class ThreeSystemDataset:
    def __len__(self) -> int:
        return 3

    def get_metadata(self, idx: int) -> tuple[int, int]:
        return (1, 0)

    def __getitem__(self, idx: int) -> tuple[AtomicData, dict[str, int]]:
        data = AtomicData(
            atomic_numbers=torch.tensor([1], dtype=torch.long),
            positions=torch.tensor([[float(idx), 0.0, 0.0]], dtype=torch.float32),
        )
        data.forces = torch.zeros(1, 3)
        data.energy = torch.zeros(1, 1)
        return data, {"idx": idx}


def main() -> None:
    # Illustrate that SizeAwareSampler stamps the system id correctly. 
    probe_sampler = SizeAwareSampler(ThreeSystemDataset(), max_atoms=2)
    initial_batch = probe_sampler.build_initial_batch()
    replacement = probe_sampler.request_replacements_budget(
        atom_budget=1,
        edge_budget=0,
        max_count=1,
    )

    print(f"initial sampler system_id: {initial_batch.system_id.view(-1).tolist()}")
    print(f"replacement sampler system_id: {replacement[0].system_id.view(-1).tolist()}")

    # Example of FusedStage pipeline losing System ID for replacement system.
    sampler = SizeAwareSampler(ThreeSystemDataset(), max_atoms=2)
    model = DemoModelWrapper(DemoModel())
    model.eval()
    fused = FusedStage(
        sub_stages=[
            (
                0,
                NVTLangevin(
                    model=model,
                    dt=0.1,
                    temperature=300.0,
                    friction=0.1,
                    random_seed=42,
                    n_steps=1,
                ),
            )
        ],
        sampler=sampler,
        refill_frequency=1,
    )

    # The first run step starts with systems 0 and 1, advances them for one
    # Langevin step, graduates them via n_steps=1, and refills with system 2.
    result = fused.run(n_steps=1)

    ids = result.system_id.view(-1).tolist()
    print(f"active system_id after one inflight refill: {ids}")

    if ids == [-1]:
        print()
        print("BUG REPRODUCED: replacement system_id was overwritten with the default.")


if __name__ == "__main__":
    main()
```

### Relevant log output

```shell
initial sampler system_id: [0, 1]
replacement sampler system_id: [2]
active system_id after one inflight refill: [-1]

BUG REPRODUCED: replacement system_id was overwritten with the default.
```

### Environment details

```shell
+ Environment location: Source checkout
+ OS: Linux
+ Python/package manager: uv
+ GPU: not required for this reproducer
```

architdatar:
## Description

Preserve sampler-assigned `system_id` values for replacement systems during inflight refill. `BaseDynamics.refill_check()` now seeds rebuilt bookkeeping tensors from values already present on the appended result batch before restoring values for systems that stayed active.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Related Issues

Fixes #112

## Changes Made

- Preserve system_id values assigned by SizeAwareSampler during `refill_check()` for inflight batching.
- Preserve system_id values for systems already in batch.
- Add test to verify this workflow.

## Testing

- [x] Unit tests pass locally (`uv run pytest test/dynamics/test_inflight.py::TestRefillCheck`)
- Added regression test for `system_id` preservation during inflight refill.

## Checklist

## Additional Notes

This is a Python-only toolkit fix. No `nvalchemi-toolkit-ops` changes are required.

### DIFF
--- a/nvalchemi/dynamics/base.py
+++ b/nvalchemi/dynamics/base.py
@@ -2047,6 +2047,17 @@ class BaseDynamics(HookRegistryMixin, _CommunicationMixin):
             device = result.device
             for key, default_fn in self._bookkeeping_keys.items():
                 new_tensor = default_fn(n_total, device)
+                # Preserve values already carried by appended replacements before
+                # restoring the prefix for systems that stayed active.
+                result_vals = getattr(result, key, None)
+                if result_vals is not None:
+                    result_vals = (
+                        result_vals.unsqueeze(-1)
+                        if result_vals.dim() == 1
+                        else result_vals
+                    )
+                    if result_vals.shape == new_tensor.shape:
+                        new_tensor.copy_(result_vals)
                 remaining_vals = getattr(batch, key, None)
                 if remaining_vals is not None and n_remaining > 0:
                     src = remaining_vals[remaining_indices]
--- a/nvalchemi/dynamics/base.py
+++ b/nvalchemi/dynamics/base.py
@@ -2058,6 +2058,11 @@ class BaseDynamics(HookRegistryMixin, _CommunicationMixin):
                     )
                     if result_vals.shape == new_tensor.shape:
                         new_tensor.copy_(result_vals)
+                    else:
+                        raise RuntimeError(
+                            f"Bookkeeping key '{key}' has shape {result_vals.shape} "
+                            f"after refill, expected {new_tensor.shape}."
+                        )
                 remaining_vals = getattr(batch, key, None)
                 if remaining_vals is not None and n_remaining > 0:
                     src = remaining_vals[remaining_indices]
--- a/test/dynamics/test_inflight.py
+++ b/test/dynamics/test_inflight.py
@@ -651,6 +651,22 @@ class TestRefillCheck:
         assert result.system_id.view(-1).tolist() == [2]
         assert result.status.view(-1).tolist() == [0]
 
+    def test_refill_preserves_mixed_system_ids(self) -> None:
+        """Remaining and replacement systems both keep their system IDs."""
+        dataset = MockDataset([(1, 0)] * 3)
+        sampler = SizeAwareSampler(dataset, max_atoms=2)
+        dynamics = BaseDynamics(model=self.model, sampler=sampler, device_type="cpu")
+
+        batch = sampler.build_initial_batch()
+        assert batch.system_id.view(-1).tolist() == [0, 1]
+
+        batch["status"] = torch.tensor([[1], [0]])
+        result = dynamics.refill_check(batch, exit_status=1)
+
+        assert result is not None
+        assert result.system_id.view(-1).tolist() == [1, 2]
+        assert result.status.view(-1).tolist() == [0, 0]
+
     def test_refill_partial_replacement(self) -> None:
         """When sampler has fewer replacements than graduated, batch shrinks.
 
--- a/test/dynamics/test_inflight.py
+++ b/test/dynamics/test_inflight.py
@@ -667,6 +667,66 @@ class TestRefillCheck:
         assert result.system_id.view(-1).tolist() == [1, 2]
         assert result.status.view(-1).tolist() == [0, 0]
 
+    def test_refill_preserves_system_ids_with_multistage_statuses(self) -> None:
+        """A status-2 system is replaced while lower-status systems remain."""
+        dataset = MockDataset([(1, 0)] * 5)
+        sampler = SizeAwareSampler(dataset, max_atoms=3)
+        dynamics = BaseDynamics(model=self.model, sampler=sampler, device_type="cpu")
+
+        batch = sampler.build_initial_batch()
+        assert batch.system_id.view(-1).tolist() == [0, 1, 2]
+
+        batch["status"] = torch.tensor([[0], [1], [2]])
+        result = dynamics.refill_check(batch, exit_status=2)
+
+        assert result is not None
+        assert result.system_id.view(-1).tolist() == [0, 1, 3]
+        assert result.status.view(-1).tolist() == [0, 1, 0]
+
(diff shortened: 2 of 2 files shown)


nvalchemi/dynamics/base.py
```diff
@@ -2047,6 +2047,17 @@ def refill_check(self, batch: Batch, exit_status: int) -> Batch | None:
...
+                # restoring the prefix for systems that stayed active.
+                result_vals = getattr(result, key, None)
+                if result_vals is not None:
+                    result_vals = (
+                        result_vals.unsqueeze(-1)
+                        if result_vals.dim() == 1
+                        else result_vals
+                    )
+                    if result_vals.shape == new_tensor.shape:
+                        new_tensor.copy_(result_vals)
```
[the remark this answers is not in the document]
architdatar: Addressed in dbcef64. Shape mismatches now raise a RuntimeError instead of silently skipping preservation.

test/dynamics/test_inflight.py
```diff
@@ -635,6 +635,22 @@ def test_refill_writes_bookkeeping_to_storage(self) -> None:
...
+
+        batch = sampler.build_initial_batch()
+        assert batch.system_id.view(-1).tolist() == [0, 1]
+
+        batch["status"] = torch.tensor([[1], [1]])
+        result = dynamics.refill_check(batch, exit_status=1)
+
+        assert result is not None
+        assert result.system_id.view(-1).tolist() == [2]
+        assert result.status.view(-1).tolist() == [0]
```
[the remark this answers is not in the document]
architdatar: Addressed in dbcef64. Added `test_refill_preserves_mixed_system_ids`, covering one remaining system plus one replacement

test/dynamics/test_inflight.py
laserkelvin, line 1: Could you add another test case where not all samples are graduating?

Essentially, we want to make sure that the refill case where only some slots are being replaced that the system ids are also being preserved then

laserkelvin (changes requested):

test/dynamics/test_inflight.py
architdatar: Thanks, I believe that this case is covered in the latest commit (`dbcef64`) by `test_refill_preserves_mixed_system_ids`.

That test starts with initial `system_id`s `[0, 1]`, marks only the first sample as graduated with `status = [[1], [0]]`, then verifies the refill result preserves the remaining system ID and the replacement ID as `[1, 2]`.

Please let me know if you have a different partial-refill scenario in mind.

laserkelvin: Basically we only cover the case where the exit and entry status codes are the default cases; you could add one or two unit test cases to make sure that this is working as intended when they're not the default 0 and 1 values

laserkelvin:
/ok to test [f7b0128](https://github.com/NVIDIA/nvalchemi-toolkit/pull/113/commits/f7b012864c57569aa3b474990c93e97a3391e22f)

test/dynamics/test_inflight.py
architdatar: Thanks for the clarification. I added a new commit, `4e7d30a` (`test: cover refill with complex status bookkeeping`), with three additional refill regression tests that cover the non-default / more complex status cases you pointed out.

The new tests cover:

- `test_refill_preserves_system_ids_with_multistage_statuses`: a multistage case with statuses `[0, 1, 2]` and `exit_status=2` (as opposed to the standard 1), where only the status-2 system graduates. This verifies the remaining lower-status systems keep their `system_id`s while the replacement gets the next sampler-assigned ID.

- `test_refill_preserves_system_ids_with_multiple_replacements`: a case where two systems graduate in the same refill call. This verifies that the remaining system keeps its ID and both replacement systems keep their newly assigned sampler IDs.

- `test_refill_preserves_replacement_status_from_dataset`: a case where replacement samples already carry nonzero graph-level status from the dataset; i.e., non-standard entry-points. This verifies that `refill_check()` preserves bookkeeping already present on appended replacement samples instead of forcing replacement status back to the default. 

Together these should cover partial refill with non-default/multistage statuses, simultaneous replacements, and replacement-carried bookkeeping values.

laserkelvin:
/ok to test [4e7d30a](https://github.com/NVIDIA/nvalchemi-toolkit/pull/113/commits/4e7d30aa52614b9f6a6f80e119e0b976229b72c6)

laserkelvin (approved):

PR #117, opened 2026-06-16
feat(models): add UMA (fairchem-core) interatomic-potential wrapper

dallasfoster:
# ALCHEMI Toolkit Pull Request

## Description

Add UMAWrapper, a BaseModelMixin-compatible wrapper around fairchem-core's
UMA (Universal Models for Atoms) MLIPPredictUnit, so UMA foundation models can
drive nvalchemi dynamics and inference. UMA is multi-task: one checkpoint
(uma-s-1p1 / uma-s-1p2 / uma-m-1p1) ships heads for OMol, OMat, OC20, ODAC,
and OMC; the wrapper pins a single task at construction (one-wrapper-one-model,
drive nvalchemi dynamics and inference. UMA is multi-task: one checkpoint
(uma-s-1p1 / uma-s-1p2 / uma-m-1p1) ships heads for OMol, OMat, OC20, ODAC,
and OMC; the wrapper pins a single task at construction (one-wrapper-one-model,
matching MACEWrapper). The conversion is tensor-native (no ASE round trip),
energy is the differentiable primitive, and forces/stress come from autograd.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Changes Made

- nvalchemi/models/uma.py: UMAWrapper (task-aware model_config, adapt_input/
  adapt_output, compute_embeddings, from_checkpoint). from_checkpoint exposes
  fairchem's native `inference_settings` (incl. "turbo"); forward routes the
  one-time lazy-init/MoLE-merge through CPU input to dodge a fairchem device-
  placement bug under turbo on GPU-resident first batches.
- nvalchemi/_optional.py + nvalchemi/models/__init__.py: register/export UMA.
- pyproject.toml: `uma` extra (fairchem-core>=2.0.0); declare uma conflicting
  with mace (e3nn pin) and cu12/cu13 (fairchem torch<2.9 vs toolkit-ops
  torch>=2.11); pin setuptools<81 for fairchem's torchtnt. uv.lock regenerated.
- examples/advanced/09_uma_nve.py: NVE/NVT/NPT MD example driven by built-in
  LoggingHook + EnergyDriftMonitorHook; NPT exercises the stress path; turbo
  selectable via inference_settings.
- test/models/test_uma.py: consolidated suite — structural (mock), forward
  equivalence vs FAIRChemCalculator, charged-input response, NVE drift (@slow),
  turbo/compile device path (@slow, CUDA).
- docs: userguide/models.md (supported-models table + UMA usage / HF-token /
[...]

PR #108, opened 2026-06-09
Adding training functionalities to Toolkit

laserkelvin:
# ALCHEMI Toolkit Pull Request

## Description

This PR introduces the core functionalities required to support training and fine-tuning of models in `nvalchemi-toolkit`.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Performance improvement
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/CD or infrastructure change

## Changes Made

- `create_model_spec` methods and dynamic pydantic model creation for `pickle`-less serialization of configuration
- Adds a few base loss functions, the general loss abstraction including individual losses and a composed loss function. The latter can be adjusted with weight scheduling, allowing the relative weighting of different losses to be adjusted over the course of training
- Adds a `TrainingStrategy` pydantic model as a recipe validation and loop executor. The execution is highly modular and extendible, allowing for (hopefully) arbitrarily complex training workflows to be built, and not limited to MLIPs
- Adds a `FineTuningStrategy` that specializes `TrainingStrategy` for...fine-tuning workflows by making pre-existing checkpoints and layer addition/modification integral to the workflow
- Adds data loading optimizations; the main changes is addition of "batched" pre-fetching, which amortizes I/O for non-contiguous data samples. This is crucial for Zarr performance when shuffling data
- Adds multidataset support, with a "meta" sampler that allows users to implement different cross-dataset sampling strategies (e.g. to account for dataset size imbalances)
- Adds several training-related hooks, such as model averaging, mixed precision, checkpointing
- Adds a CLI for training and fine-tuning: the intended use of this CLI is to provide a relatively straightforward on-ramp for users looking to get fine-tune (or train a model from scratch) quickly without needing to know the full training API

## Testing

- [x] Unit tests pass locally (`make pytest`)
- [x] Linting passes (`make lint`)
- [x] New tests added for new functionality meets coverage expectations?

## Checklist

- [x] I have read and understand the [Contributing Guidelines](../CONTRIBUTING.md)
- [x] I have updated the [CHANGELOG.md](../CHANGELOG.md)
- [x] I have performed a self-review of my code
- [x] I have added docstrings to new functions/classes
- [x] I have updated the documentation (if applicable)

## Additional Notes

> [!TIP]
> This repository uses Greptile, an AI code review service, to help conduct
[...]

PR #122, opened 2026-06-26
Add Distributed Inference and Domain Decomposition through Shard Tensor

dallasfoster:

## Description

Adds **domain decomposition (DD)** — multi-GPU spatial decomposition with halo
exchange — for nvalchemi MLIP inference and molecular dynamics, and integrates it
with current `main`. A model author opts in purely via a declarative
`distribution_spec`; `DomainParallel` then shards a system across ranks, exchanges
halo (ghost) atoms, and reduces owned-aware energies/forces/stresses so multi-rank
results match single-GPU to machine precision. Works **eager and under
`torch.compile`** across MACE (scripted + cuEq), AIMNet2, UMA, DFTD3, PME, Ewald
(incl. slab correction), and LJ, plus composed multi-model pipelines, and is
validated end-to-end through NVT and NPT MD.

## Type of Change

- [x] Performance improvement (multi-GPU scaling for large
- [ ] Documentation update

## Changes Made

- **DD runtime** (`nvalchemi/distributed/`): `DomainParalle
  spatial partitioner, `ShardedBatch`, halo storage/exchange, owned-aware reductions,
  and a single-machine distribution **validator**.
- **Declarative spec API**: `MLIPSpec`/`DistributionSpec` with `OpAdapter`/
  `MethodAdapter` (owned-slice + all-reduce), `CompilePolicy`, `GraphPadder` family,
  and `node_energy_key` so the framework reduces per-node e
- **Per-model wiring**: MACE (scripted + cuEq), AIMNet2, UMA, DFTD3, PME, Ewald, LJ
  each declare a `distribution_spec`; halo-only production path, hook-free.
- **`torch.compile` under DD**: fixed-shape caps + dead-row padding + a compile-refresh
  graph pass for zero steady-state recompiles in MD.
- **DD-correct electrostatics**: PME + Ewald, including a new moment-based
  **slab correction** (`_ops/electrostatics/slab.py`) that
  reduced — verified DD-correct.
- **Composed-model DD** (`DistributedPipelineModel`): direc
  and wired-charge (AIMNet2→PME) tiers, eager and compiled.
- **Vendored ShardTensor** backend snapshot under `distributed/_core/_upstream` to
  unblock `torch.compile`.
- **Merge of `main`**: UMA wrapper (#117), training subsystem + reporting hooks (#108),
  toolkit-ops 0.4.0 (#120), inflight-refill fix (#113); tensile-positive stress
  convention and the `nvalchemi.hooks` namespace adopted th
- Benchmarks + distributed user-guide docs/examples.

## Testing

Validated on a 2× RTX A6000 box (`nvalchemiops` 0.4.0, `NCCL_P2P_DISABLE=1`):

- **DD gate** (`test/distributed`): **612 passed / 10 failed** — all core force/energy
  equivalence (PME, Ewald, AIMNet2, MACE-scripted, pipeline, slab) **passes**; the 10
  are peripheral (see Additional Notes).
- **Slab DD correctness** (new `test_slab_multigpu.py`): world0-vs-world2
  **ΔE = 0, max|ΔF| ≈ 3.5e-7** (fp32 floor) for both PME and Ewald.
- **NVT 100-step** (MACE+DFTD3, AIMNet2+PME), world0 **and** world2: stable, all-finite,
  multi-rank trajectory tracks single-GPU.
- **NPT 100-step** (world0): stable, all-finite, completes.
- **Upstream suites green**: `test/data` (803) and `test/training` (845) fully pass —
  the merge does not regress main.
- [ ] Unit tests pass locally (`make pytest`)  *(see known test-debt below)*
- [x] Linting passes (`make lint`) on changed source
- [x] New tests added for new functionality (DD gates, slab

## Checklist

- [x] I have read and understand the Contributing Guideline
- [ ] I have updated the CHANGELOG.md
- [x] I have updated the documentation (distributed user gu

## Additional Notes

> [!TIP]


examples/distributed/03_mace_nvt_distributed.py
```diff
@@ -0,0 +1,368 @@
...
+        "--backend",
+        default="nccl",
+        help="Process-group backend. Use 'gloo' on CPU-only envs.",
+    )
+    args = parser.parse_args()
+
+    # ----- Process group setup -----
+    # ``torchrun`` populates RANK / WORLD_SIZE / LOCAL_RANK; we just
+    # bind to the assigned device and init the process group.
+    dist.init_process_group(backend=args.backend)
```
laserkelvin: Do you want to make this more `DistributedManager` oriented instead?

examples/distributed/03_mace_nvt_distributed.py
```diff
@@ -0,0 +1,368 @@
...
+    rank = dist.get_rank()
+    world_size = dist.get_world_size()
+
+    # Gloo lacks ``all_to_all`` / ``all_to_all_v``; the halo path
+    # routes through physicsnemo's ``indexed_all_to_all_v_wrapper``.
+    # Reuse the validator's isend/irecv shim (with cuda↔cpu staging)
+    # so single-GPU multi-rank correctness testing works end-to-end on
+    # Gloo. NCCL on a real multi-GPU cluster goes through the unmodified
+    # collective.
+    if args.backend == "gloo":
```
laserkelvin: Should we even have `gloo` as a testable backend for this, and not just use `nccl` as the backend?

examples/distributed/03_mace_nvt_distributed.py
```diff
@@ -0,0 +1,368 @@
...
+        backend_override=backend_override,
+    )
+
+    # ----- Load MACE wrapper -----
+    # Suppress mace-torch's chatty deprecation warnings at import.
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        from nvalchemi.models.mace import MACEWrapper
+
+    dtype = torch.float32
```
laserkelvin: Maybe have these declared at the top, if you intended for people to be able to change the `dtype`

examples/distributed/03_mace_nvt_distributed.py
```diff
@@ -0,0 +1,368 @@
...
+    if rank == 0:
+        n_frames = write_trajectory_xyz(trajectory_sink, args.output_xyz)
+        logger.info(
+            "Done. Wrote {f} xyz frames to {p}.",
+            f=n_frames,
+            p=args.output_xyz,
+        )
+
+    dynamics.close()
+    dist.destroy_process_group()
```
laserkelvin: Not sure if you feel like it's valuable, but I think it'd be pretty cool to enclose the distributed semantics into either `DomainConfig` or `DomainParallel`, so that you could then have

```python
dynamics = DomainParallel(...)

with dynamics:
    # context manager handles initialization and teardown of distributed processes
    dynamics.run(owned_batch)
```

examples/distributed/03_mace_nvt_distributed.py
laserkelvin, line 1: Should add a skip if `NVALCHEMI_SPHINX_BUILD=1` env variable check so that the gallery doesn't try and run this

examples/distributed/04_byo_pytorch_mpnn.py
```diff
@@ -0,0 +1,402 @@
...
+    def __init__(self, feat_dim: int, cutoff: float) -> None:
+        super().__init__()
+        self.cutoff = cutoff
+        centers = torch.linspace(0.0, cutoff, feat_dim)
+        self.register_buffer("centers", centers, persistent=True)
+        self.gamma = 0.5 / ((cutoff / feat_dim) ** 2)
+
+    def forward(
+        self,
+        positions: torch.Tensor,  # (n_atoms, 3)
```
laserkelvin, line 97: Not essential, but could just use `jaxtyping` or the Toolkit type hints

examples/distributed/04_byo_pytorch_mpnn.py
```diff
@@ -0,0 +1,402 @@
...
+        self.element_embedding = nn.Embedding(120, 8)
+        self.head = nn.Sequential(
+            nn.Linear(feat_dim + 8, feat_dim),
+            nn.SiLU(),
+            nn.Linear(feat_dim, 1),
+        )
+
+    def forward(
+        self,
+        positions: torch.Tensor,
```
laserkelvin, line 140: Shape hints

examples/distributed/04_byo_pytorch_mpnn.py
```diff
@@ -0,0 +1,402 @@
...
+    compile_model=True)``). The optional ``NVALCHEMI_BP_COMPILE_BACKEND``
+    overrides the backend (``inductor`` default; ``aot_eager`` for an
+    Inductor-codegen-free run). Env-var-driven so the flag propagates
+    cleanly through the validator's ``mp.spawn`` workers.
+    """
+    import os
+
+    torch.manual_seed(123)
+    model = BPModel(feat_dim=32, cutoff=5.0)
+    if os.environ.get("NVALCHEMI_BP_COMPILE", "0") != "0":
```
laserkelvin, line 304: I would rather we do this as a static variable that people could change at the top of the script, or argparse as opposed to an environment variable

examples/distributed/05_byo_graph_transformer.py
```diff
@@ -0,0 +1,739 @@
...
+        # DistributionSpec carries the chemistry-free pieces: storage layout
+        # (``policy``) and per-op distribution declarations
+        # (``custom_ops``). A non-MLIP user of the same machinery would
+        # still need this; everything chemistry-aware lives one layer
+        # up on the MLIPSpec.
+        distribution=DistributionSpec(
+            # ``policy`` — the per-field storage policy. The shipped
+            # policies (all from ``nvalchemi.distributed.ops``):
+            #
+            # * ``HaloStoragePolicy`` — each rank stores ``[owned | halo]``
```
laserkelvin, line 484: This is pretty hard to read

examples/distributed/05_byo_graph_transformer.py
laserkelvin, line 1: I think the spec generation example has value, but I kind of wanted to see this actually run with the `DomainParallel` to show people they can actually BYO

nvalchemi/distributed/domain_parallel.py
```diff
@@ -0,0 +1,667 @@
...
+        """Wrap atom positions back into the periodic box (fractional
+        coordinates modulo 1 along any PBC dim)."""
+        cell = getattr(batch, "cell", None)
+        pbc = getattr(batch, "pbc", None)
+        if cell is None or pbc is None or not pbc.any():
+            return
+
+        cell_3x3 = cell.squeeze(0)
+        inv_cell = torch.linalg.inv(cell_3x3)
+        frac = batch.positions @ inv_cell.T
```
atulcthakur: The codebase uses `cart = frac @ cell`, so this should multiply by `inv_cell`, not `inv_cell.T`. The current code wraps atoms incorrectly in skewed or triclinic cells. For non-orthogonal cells this changes positions even when they are already inside the box. Please add a skew-cell regression test after the fix.

nvalchemi/distributed/domain_parallel.py
```diff
@@ -0,0 +1,667 @@
...
+
+        if not self._forces_primed:
+            self._prime_forces(batch)
+            self._forces_primed = True
+
+        # 1. Outer BEFORE_STEP hooks.
+        self._call_hooks(DynamicsStage.BEFORE_STEP, batch)
+
+        dyn = self._dynamics
+        dyn._ensure_state_initialized(batch)
```
atulcthakur, line 274: NVE and Langevin can operate on local atom state, but Nosé–Hoover, NPH, and NPT depend on global kinetic energy, degrees of freedom, pressure, and shared cell state. This implementation computes those values independently from each rank’s shard, so ranks can evolve different thermostat/barostat states and cells. Please synchronize these quantities across the domain mesh or other less optimal option is to explicitly limit DomainParallel to NVE/Langevin.

nvalchemi/distributed/domain_parallel.py
```diff
@@ -0,0 +1,667 @@
...
+            natural,
+        ).to(torch.int64)
+        mesh = self._config.mesh
+
+        # Reshard each per-atom field independently (preserves dtypes).
+        fields: dict[str, torch.Tensor] = {"positions": batch.positions}
+        for name in ("atomic_numbers", "atomic_masses", "velocities", "forces"):
+            val = getattr(batch, name, None)
+            if val is not None:
+                fields[name] = val
```
atulcthakur: Initial partitioning supports arbitrary atom fields, but migration only moves this fixed list. Atomic charges and potentially any other custom fields can disappear after an atom crosses ranks; please migrate every existing per-atom field.

nvalchemi/distributed/domain_parallel.py
```diff
@@ -0,0 +1,667 @@
...
+        # CPU↔GPU sync at end-of-step.
+        self._dispatch_async_migrate_check(batch)
+
+        # 9. Outer AFTER_STEP hooks.
+        self._call_hooks(DynamicsStage.AFTER_STEP, batch)
+
+        self.step_count += 1
+        dyn.step_count += 1
+
+        converged = dyn._check_convergence(batch)
```
atulcthakur, line 326: Convergence is checked only against the atoms on this rank. Different ranks can therefore disagree and enter different control flow; please make convergence a mesh-wide decision.

nvalchemi/distributed/domain_parallel.py
```diff
@@ -0,0 +1,667 @@
...
+                continue
+
+            if self.step_count % hook.frequency != 0:
+                continue
+
+            scope = getattr(hook, "scope", HookScope.LOCAL)
+
+            if scope == HookScope.GLOBAL:
+                if dist.is_initialized() and getattr(batch, "energy", None) is not None:
+                    dist.all_reduce(batch.energy, op=dist.ReduceOp.SUM)
```
atulcthakur: `HookScope.GLOBAL` is documented to give hooks the complete batch, but this still passes each rank’s local shard. It also sums batch.energy, which is normally already globally reduced, so the reported energy can be multiplied by the number of ranks. Please gather the full batch over the domain mesh and avoid reducing already-global outputs again.

nvalchemi/distributed/_core/shard_tensor.py
```diff
@@ -0,0 +1,2196 @@
...
+        return _make_handler_output(
+            torch.index_select(input_plain, dim, index), input_t
+        )
+
+    # Coord is halo-exchanged at setup with PBC shifts; re-fetching would
+    # drop the shift correction. Skip the (N, 3) signature.
+    if input_plain.ndim == 2 and input_plain.shape[-1] == 3:
+        return _make_handler_output(
+            torch.index_select(input_plain, dim, index), input_t
+        )
```
atulcthakur, line 923: [P1] 
Shape alone cannot identify positions—ordinary vector features can also have shape (N, 3). Those features currently skip halo refresh and may use stale values; Would be good to directly track coordinate tensors.

nvalchemi/distributed/config.py
```diff
@@ -0,0 +1,129 @@
...
+        Submodule-name substrings to skip during ``"auto"`` discovery — for a
+        scripted op that genuinely needs cross-rank data (where marshalling to
+        local would silently give wrong numbers) or is handled via ``custom_ops``.
+    """
+
+    model_config = {"arbitrary_types_allowed": True}
+
+    cutoff: float
+    skin: float = 0.0
+    ghost_width: float | None = None
```
atulcthakur: An explicit `ghost_width` can be smaller than cutoff, which silently omits required neighbors. Please reject this configuration during setup.

nvalchemi/distributed/partitioner.py
```diff
@@ -0,0 +1,484 @@
...
+        total_cells = (
+            self.cells_per_dim[0] * self.cells_per_dim[1] * self.cells_per_dim[2]
+        )
+        if total_cells < self.world_size:
+            self.cells_per_dim = SpatialPartitioner.refine_grid_for_ranks(
+                self.cells_per_dim, self.world_size
+            )
+
+        # Compute the rank grid (Px, Py, Pz).
+        self.rank_grid: tuple[int, int, int] = SpatialPartitioner.compute_rank_grid(
```
atulcthakur, line 84: Halo exchange only contacts immediately adjacent ranks. If a rank’s domain is narrower than the interaction range, required atoms can be more than one rank away and will be missed. We can either reject these layouts or support multi-hop exchange.

nvalchemi/distributed/sharded_batch.py
```diff
@@ -0,0 +1,698 @@
...
+    """
+    if src_dtype is not None and src_dtype not in _FLOAT_DTYPE_TO_CODE:
+        raise ValueError(
+            f"ShardedBatch positions dtype {src_dtype} not supported; "
+            f"must be one of {_FLOAT_DTYPE_CODES}."
+        )
+    code = _FLOAT_DTYPE_TO_CODE.get(src_dtype, 0) if src_dtype else 0
+    code_t = torch.tensor([code], dtype=torch.int32, device=device)
+    if dist.is_initialized():
+        dist.broadcast(code_t, src=src)
```
atulcthakur: `ShardedBatch` accepts a `DeviceMesh` that may be a subgroup, but these broadcasts use the default world group and interpret the mesh-local src as a global rank. 
Subgroup meshes can therefore hang or receive data from the wrong source. Please either use the mesh process group with the correct global source rank, or explicitly reject meshes that do not cover the full world.

nvalchemi/distributed/spec.py
```diff
@@ -0,0 +1,609 @@
...
+# ``reduce=`` (e.g. UMA stress: ``OutputSpec(OutputKind.PER_GRAPH, Reduce.ALL_REDUCE)``).
+_STANDARD_MLIP_OUTPUTS: dict[str, OutputSpec] = {
+    "energy": OutputSpec(OutputKind.PER_GRAPH),
+    "forces": OutputSpec(OutputKind.PER_NODE),
+    "stress": OutputSpec(OutputKind.PER_GRAPH),
+    "atomic_energies": OutputSpec(OutputKind.PER_NODE),
+}
+
+
+SPEC_MPNN_HALO = MLIPSpec(
```
atulcthakur, line 601: `MACE` and `AIMNet2` inherit this stress definition without a final all-reduce. Each rank computes only its contribution to the strain derivative, so its stress is not the stress of the full system. The reported pressure therefore depends on how atoms are partitioned, and NPT/NPH will evolve the cell incorrectly. Please sum the stress contributions across the domain ranks.

nvalchemi/models/lj.py
```diff
@@ -139,49 +139,64 @@ def __init__(
...
+            The halo spec plus one :class:`OpAdapter` per LJ kernel.
+        """
+        import dataclasses
+
+        from nvalchemi.distributed.spec import SPEC_LJ_HALO, OpAdapter
+
+        custom_ops = (
+            OpAdapter(op=torch.ops.nvalchemi.lj_energy_forces_batch),
+            OpAdapter(op=torch.ops.nvalchemi.lj_energy_forces_virial_batch),
+        )
```
atulcthakur, line 178: The LJ kernel produces a virial contribution on each rank, but this adapter returns it without combining the ranks. Each rank therefore reports stress from only part of the system. Please accumulate virial from owned interactions without double-counting halo atoms, then sum it across the domain ranks.

nvalchemi/models/dftd3.py
```diff
@@ -539,6 +540,34 @@ def __init__(
...
+
+        from nvalchemi.distributed.spec import SPEC_DFTD3_HALO  # noqa: PLC0415
+
+        return dataclasses.replace(
+            SPEC_DFTD3_HALO,
+            distribution=dataclasses.replace(
+                SPEC_DFTD3_HALO.distribution, shard_fields=()
+            ),
+            node_energy_key="atomic_energies",
+        )
```
atulcthakur, line 573: `node_energy_key` makes the DFTD3 energy global, but it does not make the virial global. Each rank still converts only its local virial into stress. This produces incorrect pressure in NPT/NPH simulations. Please accumulate a ghost-safe local virial and sum it across ranks.

nvalchemi/distributed/distributed_model.py
```diff
@@ -0,0 +1,1022 @@
...
+        # Run energy-only through the compiled region, restoring the wrapper's
+        # active_outputs afterward.
+        mc = self._wrapper.model_config
+        saved_active = mc.active_outputs
+        mc.active_outputs = {energy_key}
+        try:
+            out = self._dd_compiled_region()(padded_batch)
+        finally:
+            mc.active_outputs = saved_active
+        e = out[energy_key]
```
atulcthakur, line 1493: This temporarily changes active_outputs to energy only, and the compiled path later constructs only energy and forces. If the caller requested stress, atomic energies, charges, or spin charges, those outputs simply disappear. For example, compiled NPT cannot receive the stress it requested. Please preserve supported outputs or raise a clear error before running an unsupported combination.

nvalchemi/distributed/spec.py
```diff
@@ -0,0 +1,609 @@
...
+halo-padded inputs follow the standard halo path; charge spreading gets an
+owned-slice + all-reduce handler. Post-spread stages (FFT, Green's function,
+IFFT, spline_gather, corrections) are replicated across ranks — they operate on
+the all-reduced global mesh, so no dispatch is needed. Caveat: single-system
+halo (``batch_idx=None``) hits a plain ``charges.sum()`` in
+``pme_energy_corrections`` that double-counts halo rows; batched halo works
+correctly via ``scatter_add_`` dispatch through :func:`per_system_reduce`."""
+
+
+SPEC_DFTD3_HALO = MLIPSpec(
```
atulcthakur, line 711: The DFTD3 kernel uses each halo neighbor’s coordination number and `dE/dCN` when calculating forces on owned atoms. Those values depend on neighbors that may lie outside a cutoff-wide halo, so `DomainConfig(cutoff=model.cutoff)` can give incomplete boundary forces. The tests already add an extra CN skin for this reason. Please make that margin automatic, validate it, or exchange the CN intermediates between ranks.

nvalchemi/distributed/sharded_batch.py
```diff
@@ -0,0 +1,698 @@
...
+            device = torch.device("cpu")
+
+        # --- Broadcast positions dtype first so cell can match ---
+        pos_dtype = _broadcast_float_dtype(
+            batch.positions.dtype if batch is not None else None,
+            device=device,
+            src=src,
+        )
+
+        # --- Broadcast cell + pbc + n_global from src ---
```
atulcthakur, line 364: This only copies cell and pbc as per-system data. Inputs such as AIMNet2 charge and mult, or UMA charge and spin, are not included when the local batch is rebuilt. Charged and open-shell systems can therefore be evaluated as neutral closed-shell systems. Please preserve and replicate all graph-level input fields.

nvalchemi/models/uma.py
```diff
@@ -332,13 +827,126 @@ def cutoff(self) -> float:
...
+            ElementReferences,
+        )
+
+        from nvalchemi.distributed.spec import (  # noqa: PLC0415
+            SPEC_UMA_HALO,
+            MethodAdapter,
+            PythonAdapter,
+        )
+
+        helpers = (
```
atulcthakur, line 1166: For OMol configurations with charge- or spin-balanced channels, balancing must be done over the full molecule. This adapter list does not replace fairchem’s `balance_channels_batched`, so every rank balances its own owned-plus-halo rows independently. The combined result can have the wrong total charge or spin and becomes dependent on the partition. Please balance globally over owned atoms and then refresh the halo values.

nvalchemi/distributed/validate/payloads.py
```diff
@@ -0,0 +1,167 @@
...
+            # reconcile (rank slice without concat) — irrecoverable.
+            if v64.shape[1:] != gv64.shape[1:]:
+                abs_diff[k] = float("inf")
+                rel_diff[k] = float("inf")
+                continue
+            elem_diff = float("inf")
+        else:
+            elem_diff = float((v64 - gv64).abs().max().item())
+
+        ref_flat_sorted = v64.flatten().sort().values
```
atulcthakur: Sorting the fully flattened output keeps only the set of numbers and loses which atom and vector component each number belonged to. A force array with values assigned to the wrong atoms or axes can therefore compare equal and make validation pass. Please match outputs using stable atom IDs and preserve the vector dimensions during comparison.

nvalchemi/distributed/_core/adapter.py
```diff
@@ -0,0 +1,1384 @@
...
+                logger.error(
+                    "AdapterRegistry.install failed for %s: %s",
+                    adapter._target_str(),
+                    e,
+                )
+                # Roll back any earlier successful installs and re-raise.
+                self.restore()
+                raise
+
+    def restore(self) -> None:
```
atulcthakur, line 1387: Two distributed models can temporarily replace the same global helper. If the older model closes first, it restores the original function while the newer model is still active; when the newer model closes, it can restore the older replacement and leave it installed globally. This is reachable in persistent model pipelines that close models in forward order. Please track replacements per target using a shared stack or reference count.

examples/distributed/03_mace_nvt_distributed.py
```diff
@@ -0,0 +1,368 @@
...
+    )
+
+    # SnapshotHook fires at AFTER_STEP and writes the resolved batch
+    # state to a DataSink. We use HostMemory for tutorial simplicity:
+    # cheap, in-memory, and we read it from rank 0 at the end of the
+    # run to write the xyz trajectory. For longer runs swap in
+    # ZarrData for incremental disk persistence.
+    n_frames_expected = (args.n_steps // args.snapshot_every) + 1
+    trajectory_sink = HostMemory(capacity=n_frames_expected)
+    snapshot_hook = SnapshotHook(
```
atulcthakur, line 274: Could we make this a rank-zero hook that gathers the full batch? `SnapshotHook` has no scope, so `DomainParallel` defaults it to local and writes only each rank’s owned atoms. Since the example later saves only rank 0’s sink, the resulting XYZ contains only rank 0’s shard on a real multi-rank run.

examples/distributed/03_mace_nvt_distributed.py
```diff
@@ -0,0 +1,368 @@
...
+from nvalchemi.data import AtomicData, Batch
+from nvalchemi.distributed import DomainConfig, DomainParallel
+from nvalchemi.dynamics import HostMemory, NVTLangevin
+from nvalchemi.dynamics.base import DynamicsStage
+from nvalchemi.dynamics.hooks import SnapshotHook
+from nvalchemi.hooks import NeighborListHook
+
+# Reuse the SiO2 supercell builder from the benchmark suite — one
+# canonical periodic test system across the distributed examples.
+sys.path.insert(0, str(Path(__file__).parent))
```
atulcthakur: `_benchmark_common.py` only exists under `benchmark/distributed`, but this adds `examples/distributed` to the import path. The documented command therefore fails here with `ModuleNotFoundError`. Please move the shared helper into an importable module or import it from the correct location.

examples/distributed/03_mace_nvt_distributed.py
```diff
@@ -0,0 +1,368 @@
...
+            "Done. Wrote {f} xyz frames to {p}.",
+            f=n_frames,
+            p=args.output_xyz,
+        )
+
+    dynamics.close()
+    dist.destroy_process_group()
+
+
+if __name__ == "__main__":
```
atulcthakur, line 347: Once the import above is fixed, the default Sphinx Gallery build will execute this example as `__main__`. Example 03 ignores `NVALCHEMI_SPHINX_BUILD` and calls `init_process_group() `without a torchrun environment, so the docs build will fail. Please add the same guard used by examples 01 and 02.

atulcthakur (changes requested):

nvalchemi/distributed/strategy.py
```diff
@@ -0,0 +1,1079 @@
...
+    # Order matters: GraphParallelPolicy / GraphReplicatePolicy subclass
+    # PlainShard, HaloStoragePolicy is standalone (RefreshOnlyHaloPolicy
+    # subclasses it and is also halo).
+    if isinstance(policy, GraphReplicatePolicy):
+        return GraphReplicateStrategy(policy, config, rank)
+    if isinstance(policy, GraphParallelPolicy):
+        return GraphPartitionStrategy(policy, config, rank)
+    if isinstance(policy, HaloStoragePolicy):
+        return HaloStrategy(policy, config, rank)
+    raise ValueError(
```
laserkelvin, line 970: If I understand correctly, the rest of the design for the policy abstraction allows for user defined ones, but this factory method blocks that since it doesn't check against a registry or something like that

nvalchemi/distributed/spec.py
```diff
@@ -0,0 +1,648 @@
...
+    # scatter + all-reduce). For electrostatics/dispersion wrappers (PME, Ewald,
+    # DFTD3) whose per-atom energies are plain tensors that can't route through
+    # ``per_system_reduce``: they emit raw per-atom energies under this key plus
+    # a plain ``"energy"`` for the non-distributed case, and the framework
+    # overrides ``"energy"`` with the owned-aware sum under decomposition.
+    # ``None`` (default) means the wrapper owns its own per-system energy. The
+    # compiled path does the same via
+    # ``CompilePolicy.force_strategy=FRAMEWORK_FROM_NODE_ENERGY``.
+    node_energy_key: str | None = None
+    # ``outputs`` is the recommended declaration form (lowered onto the canonical
```
laserkelvin, line 379: If possible, these should be embedded in `Annotated` descriptions, instead of having then as just inline comments

nvalchemi/distributed/spec.py
```diff
@@ -0,0 +1,648 @@
...
+                self, "owned_only_outputs", self.owned_only_outputs | owned
+            )
+            object.__setattr__(
+                self, "all_reduce_outputs", self.all_reduce_outputs | all_reduce
+            )
+            object.__setattr__(self, "output_kinds", kinds)
+            # Clear ``outputs`` once consumed: the canonical fields are the truth.
+            object.__setattr__(self, "outputs", None)
+
+    def merge(self, other: "MLIPSpec") -> "MLIPSpec":
```
laserkelvin, line 418: Can't this be actually typed, instead of having string types for `MLIPSpec`?

nvalchemi/distributed/spec.py
```diff
@@ -0,0 +1,648 @@
...
+                self.distribution.third_party_helpers + other.distribution.third_party_helpers
+            ),
+            # Union: the composed model promotes whatever either side promotes.
+            shard_fields=tuple(
+                dict.fromkeys(
+                    self.distribution.shard_fields + other.distribution.shard_fields
+                )
+            ),
+        )
+        return MLIPSpec(
```
laserkelvin, line 445: What about compliation policy?

nvalchemi/distributed/spec.py
```diff
@@ -0,0 +1,648 @@
...
+    from it.
+
+    The policy is only the contract. Whether and how to compile is owned by
+    :class:`DistributedModel` (constructed with ``compile=True`` /
+    ``compile_kwargs=...``); the policy carries no compile switch of its own.
+    """
+
+    static_shapes: bool = True
+    graph_padder: "GraphPadder | None" = None
+    force_strategy: ForceStrategy = ForceStrategy.MODEL_INTERNAL
```
laserkelvin, line 261: Is this really part of `CompliePolicy`, or should it belong with `MLIPSpec`?

nvalchemi/distributed/config.py
```diff
@@ -0,0 +1,155 @@
...
+    strategy: StrategyKind = StrategyKind.HALO
+    mesh: Any = None  # DeviceMesh at runtime
+    mesh_dim: str = "domain"
```
laserkelvin, line 124: Type as `Literal` or enum

nvalchemi/distributed/strategy.py
```diff
@@ -0,0 +1,1079 @@
...
+    # ---- reductions (intent → mechanism) --------------------------------
+
+    def _group(self) -> Any:
+        """Process group this strategy's collectives run on.
+
+        Confined to the mesh; per the mesh-dim contract this should be the named
+        sub-mesh group (``config.mesh_dim``), but the 1-D default group is
+        equivalent today and the named-dim cutover is a later step.
+        """
+        return mesh_group(self._config.mesh)
```
laserkelvin: What happens to `mesh_dim`?

nvalchemi/distributed/domain_parallel.py
```diff
@@ -0,0 +1,631 @@
...
+                raise ValueError("batch must be provided in single-process mode")
+            self._n_owned = batch.positions.shape[0]
+            return batch
+
+        mesh = self._config.mesh
+
+        # Adapter around the inner dynamics' model. Owns halo exchange,
+        # NL rebuild, and output consolidation. Built first so the strategy can
+        # be selected from its resolved storage policy.
+        self._dist_model = DistributedModel(self._dynamics.model, self._config)
```
laserkelvin: `compile` flag(s) isn't passed here into `DistributedModel`; do we need to include that in the `DomainConfig`?

nvalchemi/distributed/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+    if name in _imports:
+        module_path, attr = _imports[name]
+        import importlib
+
+        module = importlib.import_module(module_path)
+        return getattr(module, attr)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+__all__ = [
```
laserkelvin, line 209: This doesn't include `DistributedModel` and `DistributedPipelineModel` - is that intentional?

nvalchemi/distributed/distributed_pipeline.py
```diff
@@ -0,0 +1,440 @@
...
+        consumer = plan["consumer"].model
+        out_key = plan["producer_out_key"]
+        field = plan["field"]
+
+        prod_mc = producer.model_config
+        saved_prod = prod_mc.active_outputs
+        prod_mc.active_outputs = {"energy", out_key}
+        try:
+            sharded.invalidate_padded_view()
+            with DistributedModel(producer, plan["producer_cfg"]) as pdm:
```
laserkelvin, line 350: Drops compilation?

nvalchemi/distributed/distributed_pipeline.py
```diff
@@ -0,0 +1,440 @@
...
+
+                # Consumer: direct kernel force + energy differentiable in the
+                # wired field, whose ghost values are gathered (autograd-aware)
+                # from the producer's owned values.
+                sharded.invalidate_padded_view()
+                cons_mc = consumer.model_config
+                saved_cons = cons_mc.active_outputs
+                cons_mc.active_outputs = {"energy", "forces"}
+                try:
+                    with DistributedModel(consumer, plan["consumer_cfg"]) as cdm:
```
laserkelvin, line 372: Would it be worth refactoring this a little bit (it could be in a later PR, but I guess doing it now would be good too) so that we create the distributed models once and then re-use them?

nvalchemi/distributed/distributed_model.py
```diff
@@ -0,0 +1,1053 @@
...
+        """Build the halo config from the sharded batch's geometry the
+        first time we see one. When available, reuse the partitioner
+        cached on :attr:`ShardedBatch.partitioner` (built there from the
+        same config + broadcast cell/pbc) to avoid duplicate work and
+        potential drift. Fall back to constructing one from the sharded
+        batch's geometry when not available (e.g. gloo-harness batches
+        built outside :meth:`ShardedBatch.from_batch`).
+        """
+        if self._partitioner is not None:
+            return
```
laserkelvin, line 784: If I understand correctly, this will make it so that you can't re-use an initialized `DistributedModel` a second time, if we use the context manager:

```python
model = DistributedModel(...)

# first time initializing will work
with model:
    do_stuff()

# if you try to re-use the instance, things do not get set up again
# after `close()` removed them
with model:
    do_more_stuff()
```

Just wanted to confirm that this is not a supported pattern, and if not we should probably include this as like a gotcha in the documentation (users should do everything in a single context)

nvalchemi/distributed/_core/adapter.py
```diff
@@ -0,0 +1,1383 @@
...
+        )
+
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize to a JSON-roundtrippable dict."""
+        return {
+            "kind": "method",
+            "module_path": self.module_path,
+            "class_name": self.class_name,
+            "method_name": self.method_name,
+            "replacement": _replacement_qualname(self.replacement),
```
laserkelvin, line 1121: Drops `mode` from serialization

nvalchemi/distributed/config.py
```diff
@@ -0,0 +1,155 @@
...
+        all. Overridable via ``NVALCHEMI_SCRIPTED_MARSHAL``.
+    scripted_marshal_exclude : tuple[str, ...]
+        Submodule-name substrings to skip during ``"auto"`` discovery — for a
+        scripted op that genuinely needs cross-rank data (where marshalling to
+        local would silently give wrong numbers) or is handled via ``custom_ops``.
+    """
+
+    model_config = {"arbitrary_types_allowed": True}
+
+    cutoff: float
```
laserkelvin: Since this is a pydantic model, you should add constraints like making sure these values are valid, e.g. `cutoff > 0`

nvalchemi/distributed/config.py
```diff
@@ -0,0 +1,155 @@
...
+    skin: float = 0.0
+    ghost_width: float | None = None
+    strategy: StrategyKind = StrategyKind.HALO
+    mesh: Any = None  # DeviceMesh at runtime
```
laserkelvin: Can't you type it with `DeviceMesh | None` then?

nvalchemi/distributed/domain_parallel.py
```diff
@@ -0,0 +1,631 @@
...
+
+    # ------------------------------------------------------------------
+    # Run
+    # ------------------------------------------------------------------
+
+    def run(self, batch: Batch, n_steps: int | None = None) -> Batch:
+        """Run the domain-decomposed simulation for *n_steps* steps."""
+        # Single-process fallback — delegate to the inner dynamics' run.
+        if self._dist_model is None:
+            return self._dynamics.run(batch, n_steps=n_steps)
```
laserkelvin, line 665: Is this just to support running `DomainParallel` on a single GPU?

nvalchemi/models/base.py
```diff
@@ -523,6 +531,60 @@ def add_output_head(self, prefix: str) -> None:
         """
         raise NotImplementedError
 
+    # ------------------------------------------------------------------
+    # Distributed hooks
+    # ------------------------------------------------------------------
+
+    def distribution_spec(self, strategy: Any = None) -> Any:
```
laserkelvin: Return type should be `MLIPSpec` - propagate to child models

nvalchemi/distributed/_core/collection.py
laserkelvin, line 1: Centralize the communications methods like `world_size()`

nvalchemi/distributed/_core/collection.py
```diff
@@ -0,0 +1,300 @@
...
+    """Broadcast a full ``(n_global, *trailing)`` tensor from *src* to all ranks.
+
+    Broadcast is collective on every backend (NCCL, gloo) and needs no per-rank
+    P2P; for a one-time setup scatter the ``n_global x world_size`` bandwidth vs
+    a perfect scatter is negligible. Each rank then slices its own rows per its
+    field policy.
+    """
+    full_shape = (n_global,) + trailing
+    if local_rank == src:
+        assert src_tensor is not None  # noqa: S101
```
laserkelvin: Should we raise here instead of using `assert` instead of suppressing the linter?

nvalchemi/distributed/_core/particle_halo.py
```diff
@@ -0,0 +1,1266 @@
...
+        send_indices_owned.append(idx_own)
+
+    # All-gather send counts → sizes matrix.
+    local_counts = torch.zeros(world_size, dtype=torch.int64, device=device)
+    for r in range(world_size):
+        local_counts[r] = send_indices[r].shape[0]
+
+    all_counts_list = [torch.zeros_like(local_counts) for _ in range(world_size)]
+    dist.all_gather(all_counts_list, local_counts, group=group)
+    sizes = [c.tolist() for c in all_counts_list]
```
laserkelvin: Does this need to be blocking? Can it not stay as a tensor?

nvalchemi/distributed/_core/particle_halo.py
```diff
@@ -0,0 +1,1266 @@
...
+def halo_scatter_correct_op(
+    padded: torch.Tensor,
+    send_idx_flat: list[int],
+    send_idx_lens: list[int],
+    send_sizes_flat: list[int],
+    n_owned: int,
+    rank: int,
+    world_size: int,
+) -> torch.Tensor:
+    """Dispatcher-visible halo correction for the compile path. Runs eagerly at
```
laserkelvin: Full docstring since this is public

nvalchemi/distributed/_core/particle_halo.py
```diff
@@ -0,0 +1,1266 @@
...
+def halo_forward_op(
+    owned: torch.Tensor,
+    send_idx_flat: list[int],
+    send_idx_lens: list[int],
+    send_sizes_flat: list[int],
+    n_padded: int,
+    rank: int,
+    world_size: int,
+) -> torch.Tensor:
+    """Owned rows -> padded ``[owned | halo]`` (gather neighbors' owned rows into
```
laserkelvin: Full docstring needed

nvalchemi/distributed/particle_halo.py
```diff
@@ -0,0 +1,176 @@
...
+)
+
+if TYPE_CHECKING:
+    from nvalchemi.distributed._core.halo_types import ParticleHaloConfig
+
+__all__ = ["halo_exchange"]
+
+
+def halo_exchange(
+    sharded: Any,
```
laserkelvin, line 49: `ShardedBatch` instead of `Any`?

nvalchemi/distributed/particle_halo.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Every per-atom field scattered onto the ShardedBatch rides through.
+    # positions has already been padded (autograd-aware); forces is
+    # reset to zero because the model writes into it in place; every
+    # other field — atomic_numbers, atomic_masses, velocities, charges,
+    # momenta, anything the producer attached via add_node_property —
+    # halo-gathers through ``pad_field``.
+    atom_fields = sharded.atom_fields()
+
+    def _build_padded_field(name: str, shard: Any) -> torch.Tensor:
```
laserkelvin, line 115: Could potentially centralize this somewhere; I think it's generally useful to have padding and unpadding utilities

nvalchemi/distributed/particle_halo.py
```diff
@@ -0,0 +1,176 @@
...
+            sharded.pbc if sharded.pbc.ndim == 2 else sharded.pbc.unsqueeze(0)
+        )
+
+    padded_data = AtomicData(**padded_kwargs)
+    for name, shard in atom_fields.items():
+        if name in ("positions", "atomic_numbers", "atomic_masses"):
+            continue
+        padded_data.add_node_property(name, _build_padded_field(name, shard))
+
+    sharded.padded_batch = BatchCls.from_data_list([padded_data], device=device)
```
laserkelvin, line 175: Add `skip_validation` flag to method so we drop the overhead

proposal-distributed-graph-parallel.md
laserkelvin, line 1: This doesn't need to be here

nvalchemi/distributed/distributed_model.py
```diff
@@ -0,0 +1,1053 @@
...
+            clear_gp_compile_routing()
+        _mark("wrapper_forward")
+
+        grp = (
+            mesh_group(mesh)
+            if (dist.is_initialized() and world > 1 and mesh is not None)
+            else None
+        )
+        # Energy: each rank holds its owned per-system partial → global SUM.
+        if "energy" in output and isinstance(output["energy"], torch.Tensor):
```
laserkelvin, line 1228: I'm wondering if it's worth fusing the energy and force reductions - the energy one is going to be latency bound anyway

nvalchemi/distributed/distributed_model.py
```diff
@@ -0,0 +1,1053 @@
...
+        # Forces: the model returns ``-dE_owned/d pos`` over the full positions.
+        # The feature all-gather's reduce-scatter backward already routed each
+        # node's gradient to its owner once, so a plain SUM (no ``/world``) is
+        # the global force. Slice to this rank's owned atoms (gathered back to
+        # the global ordering by consolidation as an owned-only output).
+        if "forces" in output and isinstance(output["forces"], torch.Tensor):
+            f = output["forces"]
+            if grp is not None:
+                f = f.clone()
+                dist.all_reduce(f, op=dist.ReduceOp.SUM, group=grp)
```
laserkelvin: Instead of `all_reduce`, you could write it as a reduce-scatter so you don't need to communicate the full tensor

examples/distributed/04_byo_pytorch_mpnn.py
```diff
@@ -0,0 +1,402 @@
...
+        atomic_energies = raw["atomic_energies"]
+
+        total_energy = torch.zeros(
+            n_graphs, device=positions.device, dtype=positions.dtype
+        )
+        total_energy = total_energy.scatter_add_(0, data.batch_idx, atomic_energies)
+        raw["energy"] = total_energy
+
+        if compute_forces:
+            (forces_grad,) = torch.autograd.grad(
```
laserkelvin, line 242: Didn't we have a helper function for this? Or is this just for transparency?

Same comment applies to the other  BYO exampl

docs/userguide/distributed_design.md
```diff
@@ -0,0 +1,1361 @@
...
+    SPEC_EWALD_HALO,       # Ewald (real + reciprocal stages)
+    SPEC_PME_HALO,         # PME (charge spread + FFT mesh)
+)
+```
+
+```{code-block} python
+:caption: The shipped presets cover the production model families.
+from nvalchemi.distributed.spec import (
+    SPEC_MPNN_HALO,        # MACE, NequIP, Allegro, ORB
+    SPEC_AIMNET2_GATHER,   # AIMNet2 (sharded, mol_sum reductions)
```
laserkelvin: I don't think this was exported in that namespace

docs/userguide/distributed_design.md
laserkelvin, line 1: I searched for `ReplicatedMetadata` symbol and couldn't find anything

laserkelvin (changes requested):
Leaving some initial comments

nvalchemi/distributed/domain_parallel.py
dallasfoster: Fixed. The fractional-coordinate wrap now uses `inv_cell` (`frac = positions @
inv_cell`) rather than the transpose, so triclinic/non-orthogonal cells wrap
correctly. Verified on a skewed cell round-trip (max error ~6e-8).

nvalchemi/distributed/domain_parallel.py
dallasfoster: Fixed. `_resolve_pending_migrate` no longer reshards a hardcoded field list — it
enumerates `batch._atoms_group` and reattaches every per-atom field generically
(known fields via setattr, custom node properties via `add_node_property`), and
also carries per-graph `stress` through the rebuild. This fixes a class of bugs
where custom int64 node properties (e.g. an atom id) or `stress` were dropped at
the first migration and caused stale-size / OOB device asserts.

nvalchemi/distributed/domain_parallel.py
dallasfoster: This was the substantive one and it's now addressed. `DomainParallel`
integrated each rank's owned shard independently, which is exact for NVE/Langevin
but wrong for Nosé–Hoover / NPT / NPH, whose controllers couple to *global*
quantities (kinetic energy, degrees of freedom, pressure) that were being
computed per-shard.

The fix reduces to two mesh-global primitives + replicated controller state, and
is implemented as a `GlobalThermoSync` coordinator wired into `DomainParallel`
(no DD verbs leak into the integrators — they stay single-process):
- **global KE** — local `Σ½mv²` reduced with `all_reduce(SUM)` over the mesh;
- **global DOF** — `ndof` (and the derived controller masses: NHC `Q`, barostat
  `W`) overwritten with the mesh-global atom count post-partition;
- **replicated controller state** — per-step rank-0 broadcast of the NHC chain /
  barostat state so ranks can't drift.

nvalchemi/distributed/domain_parallel.py
dallasfoster: Fixed. Convergence is now a mesh-wide `all_reduce(MIN)` so a rank can't declare
the system converged based only on its owned shard.

nvalchemi/distributed/validate/payloads.py
dallasfoster: Fixed for the axis-scramble failure mode: the validation sort (`_lexsort_rows`)
now preserves the vector dimensions instead of flattening, which catches
component-scrambled mismatches the flatten-then-sort missed. Full atom-identity
matching (rather than coordinate-sort) is a follow-up — noted.

nvalchemi/distributed/domain_parallel.py
dallasfoster: Fixed. `HookScope.GLOBAL` now gathers the full batch before invoking the hook and
no longer double-reduces the energy (the forward already consolidated it). The
per-rank/global scopes are handled explicitly in `DomainParallel`.

nvalchemi/distributed/_core/adapter.py
dallasfoster: Agreed this is a real hazard when adapters nest/overlap. The fix (a per-target
restore stack / refcount so restores happen in LIFO order) is designed but I've
deliberately not landed it in this PR — it's load-bearing for the adapter install
path and I want dedicated multi-GPU tests around it first. Tracking as a
follow-up.

examples/distributed/03_mace_nvt_distributed.py
dallasfoster: Fixed — the snapshot hook now runs at `RANK_ZERO` scope so it writes once from the
owning rank rather than per-shard.

examples/distributed/03_mace_nvt_distributed.py
dallasfoster: Fixed — corrected to the real `benchmark/distributed` import path.

examples/distributed/03_mace_nvt_distributed.py
dallasfoster: Fixed — the example early-returns under `NVALCHEMI_SPHINX_BUILD=1` and when no
torchrun env is present, so the gallery build doesn't execute the distributed run.

nvalchemi/distributed/distributed_model.py
dallasfoster: Confirmed single-lifecycle by design (setup installs process-wide adapter state we
don't want to double-install). Documented the contract on the class/`__enter__`
and added a guard that raises `RuntimeError("DistributedModel is single-use;
construct a new one after close()")` instead of silently running half-set-up.

nvalchemi/distributed/distributed_model.py
dallasfoster: Done. The energy all-reduce (a small, latency-bound per-system vector) is now
issued `async_op=True` and overlapped under the force reduce-scatter in both GP
paths, so we don't pay a serialized second round-trip.

nvalchemi/distributed/distributed_model.py
dallasfoster: Done. Both graph-parallel force reductions now use `reduce_scatter_tensor` into an
owned-shaped buffer (`_reduce_scatter_owned`) instead of all-reducing the full
`[N,3]` and slicing. Box-validated force-equivalent (eager + compile, incl.
uneven partitions). Honest note: on the 2-node sweep this came out perf-neutral —
the force collective wasn't the bottleneck (the per-layer feature all-gather
dominates) — but it's the correct, leaner pattern and lands as requested.

nvalchemi/distributed/strategy.py
dallasfoster: Done. Added a module-level policy→strategy registry (`register_strategy` +
MRO-walk resolve); the two built-ins register through it and the `raise` stays as
the fallback for a genuinely unregistered policy. User policies can now bind their
own strategy without editing the factory.

nvalchemi/distributed/strategy.py
dallasfoster: Done. `_group()` now resolves the named sub-mesh group via `config.mesh_dim` when
the mesh has named dims, falling back to the whole mesh otherwise — `mesh_dim` is
no longer dead config.

nvalchemi/distributed/domain_parallel.py
dallasfoster: Addressed in the current branch — `DomainConfig` carries `compile: bool` and
`partition()` threads it through (`compile=self._config.compile`); I also dropped
the defensive `getattr(..., "compile", False)` now that the field is guaranteed.

nvalchemi/distributed/__init__.py
dallasfoster: Done — both are now in the lazy-import map and `__all__` (public-API stability
test updated), so the exported surface matches the docs.

nvalchemi/distributed/distributed_pipeline.py
dallasfoster: Both are real and go together. The wired-autograd group constructs its
producer/consumer `DistributedModel`s inside every `__call__` (rebuilding +
re-installing adapters per step) and without `compile=`, unlike the per-step path
which goes through `_make_dist_model`. I'll hoist construction to plan/`__enter__`
time and route it through `_make_dist_model` (fixing both the reuse and the
dropped compile), keeping only the per-step `active_outputs`/halo bookkeeping in
`__call__`. Because it changes model lifetime I want to land it with wired
MACE+DFTD3 / AIMNet2+PME box validation, so I'm tracking it as a follow-up issue.

nvalchemi/distributed/distributed_pipeline.py
dallasfoster: Both are real and go together. The wired-autograd group constructs its
producer/consumer `DistributedModel`s inside every `__call__` (rebuilding +
re-installing adapters per step) and without `compile=`, unlike the per-step path
which goes through `_make_dist_model`. I'll hoist construction to plan/`__enter__`
time and route it through `_make_dist_model` (fixing both the reuse and the
dropped compile), keeping only the per-step `active_outputs`/halo bookkeeping in
`__call__`. Because it changes model lifetime I want to land it with wired
MACE+DFTD3 / AIMNet2+PME box validation, so I'm tracking it as a follow-up issue.

nvalchemi/distributed/config.py
dallasfoster: Done — `cutoff: Field(gt=0)`, `skin: Field(ge=0)`, `ghost_width: Field(gt=0)`,
positive `grid_dims`, so a physically meaningless config fails at construction.

benchmark/distributed/README_fairchem_native.md
```diff
@@ -0,0 +1,126 @@
+# fairchem-native distributed NVT benchmark
+
+`benchmark_fairchem_native_nvt.py` measures **fairchem's OWN graph-parallel
+MD** throughput and per-rank peak memory on the exact bcc-Fe systems and
+metrics used in `DD_SCALING_SWEEP_2026-07-01.md` §7a. It is the head-to-head
```
laserkelvin: Rogue document

nvalchemi/distributed/config.py
dallasfoster: I tried this and had to revert it to `Any`. Pydantic strict-validates the
annotation even with `arbitrary_types_allowed`, and it rejects the ducktyped mock
meshes the test harness uses (which `mesh_group` deliberately supports), breaking
`test_config`/`test_particle_halo`. Kept `Any` with a clarifying comment about the
mock-mesh contract.

nvalchemi/distributed/_core/adapter.py
dallasfoster: Done — `scripted_marshal: Literal["auto","declared","off"]`. Left `mesh_dim` as a
free-form `str` since it's a user-chosen dimension name.

nvalchemi/distributed/domain_parallel.py
dallasfoster: Yes — it's the intentional single-process / single-GPU (unlaunched) passthrough,
so the identical example/user code runs under `torchrun -n>1` and as a plain
single-GPU script (and in tests / the docs build). Sharpened the inline comment to
say so explicitly.

nvalchemi/models/base.py
dallasfoster: Base signature is now `distribution_spec(strategy: StrategyKind | None) ->
MLIPSpec | None` (TYPE_CHECKING imports). Propagating the same annotation to the
seven wrapper overrides is a mechanical sweep I'll do as a follow-up issue.

nvalchemi/distributed/_core/collection.py
dallasfoster: Done — replaced the `assert` (+ the sibling ones) with explicit `raise
ValueError`/`RuntimeError` and dropped the `# noqa: S101`, so the guards survive
`python -O`.

benchmark/distributed/fairchem_worker_peakmem.patch
laserkelvin, line 1: Is this intentional?

nvalchemi/distributed/_core/particle_halo.py
dallasfoster: Done — `halo_scatter_correct_op` and `halo_forward_op` now have full
Parameters/Returns docstrings, including a note that the int[] marker marshaling
is an inductor-lowering constraint.

nvalchemi/distributed/_core/particle_halo.py
dallasfoster: Done — halo_scatter_correct_op and halo_forward_op now have full
Parameters/Returns docstrings, including a note that the int[] marker marshaling
is an inductor-lowering constraint.

nvalchemi/distributed/particle_halo.py
dallasfoster: The exact line drifted since the snapshot, but the point stands: `pad_field` takes
a genuine `ShardTensor | Tensor` union and should be typed as that; where a param
is truly a `ShardedBatch` I'll type it via a `TYPE_CHECKING` import to dodge the
cycle. If you were pointing at a specific `sharded`-batch param, re-drop the
comment on the current line and I'll type it there.

nvalchemi/distributed/particle_halo.py
dallasfoster: Agreed — under steady-state MD the domain geometry is stable so the per-step
boundary validation is redundant after warmup. I'll add an opt-in
`skip_validation` (default off, threaded from `DomainConfig`) as a follow-up so
the default keeps the correctness checks. Not in this PR yet.

docs/userguide/distributed_design.md
dallasfoster: Done — fixed the code-block to the real preset names (`SPEC_MPNN_HALO`,
`SPEC_MPNN_GP`, `SPEC_UMA_HALO`, `SPEC_LJ_HALO`, `SPEC_EWALD_HALO`,
`SPEC_PME_HALO`, `SPEC_DFTD3_HALO`).

docs/userguide/distributed_design.md
dallasfoster: Confirmed — it's leftover from the earlier UMA-replicate design and appears only
in this doc. I'm rewriting that section to the shipped design (node-partition GP
via the fairchem `gp_utils` swap) and auditing the doc for other non-existent
symbols, as a docs follow-up issue.

benchmark/distributed/benchmark_fairchem_native_nvt.py
```diff
@@ -0,0 +1,484 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""fairchem-native distributed NVT MD benchmark — a baseline for §7a.
+
+This times fairchem's OWN graph-parallel MD throughput + per-rank peak
+memory so it can sit next to the ``UMA graph_partition`` NVT numbers in
+``DD_SCALING_SWEEP_2026-07-01.md`` §7a (p50 step_ms + per-rank peak MiB,
```
laserkelvin: Document mention

benchmark/distributed/benchmark_fairchem_native_nvt.py
```diff
@@ -0,0 +1,484 @@
...
+A CSV with the same columns as §7a's NVT CSV —
+``model,env_torch,strategy,world,n_atoms,step_ms_p50,peak_MiB_per_rank,status``
+— labeled ``fairchem_native_gp``, plus a human table on stdout. Per-size
+OOM is caught, recorded as ``status=OOM``, and the sweep continues.
+
+Launch
+------
+Single node only; Ray spawns the actors (no torchrun)::
+
+    # A6000 box (48 GB): worlds 1/2 only. UMA turbo ~13.7 MiB/atom → a
```
laserkelvin: Too much detail?

examples/distributed/06_mace_npt_distributed.py
```diff
@@ -0,0 +1,386 @@
...
+            pct=100.0 * (v1 - v0) / v0,
+            n=args.n_steps,
+        )
+        n_frames = write_trajectory_xyz(trajectory_sink, args.output_xyz)
+        logger.info(
+            "Done. Wrote {f} xyz frames to {p}.", f=n_frames, p=args.output_xyz
+        )
+
+    dynamics.close()
+    dist.destroy_process_group()
```
laserkelvin: Maybe wrap everything in a `try/except/finally` so this will always cleanup

examples/distributed/06_mace_npt_distributed.py
```diff
@@ -0,0 +1,386 @@
...
+        barostat_time=args.barostat_time_fs,
+        thermostat_time=args.thermostat_time_fs,
+        pressure_coupling="isotropic",
+        chain_length=3,
+        hooks=[nl_hook],
+        n_steps=args.n_steps,
+    )
+
+    # ----- DomainParallel wrapping (SnapshotHook on the outer, as in 03) -----
+    dynamics = DomainParallel(
```
laserkelvin: Context manager pattern

nvalchemi/distributed/distributed_model.py
```diff
@@ -57,14 +57,89 @@
 __all__ = ["DistributedModel", "DistributionError"]
 
 
+def isolate_compile_cache_per_rank() -> None:
```
laserkelvin, line 60: Not sure if it's overengineering, but might be good to scope this with as a context manager so that the environment variables are reset outside for good sanitation

nvalchemi/distributed/distributed_model.py
```diff
@@ -57,14 +57,89 @@
...
+    """Give each rank its own ``torch.compile`` on-disk cache dir (multi-rank DD).
+
+    The inductor FxGraphCache + the AOTAutograd cache default to one shared dir;
+    under multi-rank DD a rank can deserialize another rank's guarded entry and
+    raise a ``KeyError`` mid-forward → a skipped collective → NCCL deadlock.
+    Pointing each rank at its own dir removes the collision while keeping the
+    caches ON (disabling them instead re-lowers the AOT graph every step, ~26 s/step
+    on torch 2.10).
```
laserkelvin: Weird detali

nvalchemi/distributed/distributed_model.py
```diff
@@ -57,14 +57,89 @@
...
+
+    root = os.path.join(tempfile.gettempdir(), "nvalchemi_dd_compile_cache")
+    for _var, _sub in (
+        ("TORCHINDUCTOR_CACHE_DIR", "inductor"),
+        ("TRITON_CACHE_DIR", "triton"),
+    ):
+        os.environ.setdefault(_var, os.path.join(root, f"{_sub}_rank{rank}"))
+
+
+def _configure_dd_dynamo() -> None:
```
laserkelvin, line 97: And/or make this a context manager (as well)

presentation/SPEAKER_NOTES.md
laserkelvin, line 1: Probably don't need this

DD_SCALING_SWEEP_2026-07-01.md
laserkelvin, line 1: Remove?

examples/distributed/03_mace_nvt_distributed.py
dallasfoster: done.

examples/distributed/03_mace_nvt_distributed.py
dallasfoster: agreed, changed to just nccl

examples/distributed/03_mace_nvt_distributed.py
dallasfoster: done.

examples/distributed/03_mace_nvt_distributed.py
dallasfoster: done.

examples/distributed/04_byo_pytorch_mpnn.py
dallasfoster: just for transparency

benchmark/distributed/README_fairchem_native.md
dallasfoster: removed.

benchmark/distributed/benchmark_fairchem_native_nvt.py
dallasfoster: removed.

benchmark/distributed/benchmark_fairchem_native_nvt.py
dallasfoster: removed.

nvalchemi/distributed/distributed_model.py
dallasfoster: difficult to do in this situation since the compile arguments need to exist in scope as long as the model is runnable.

nvalchemi/distributed/distributed_model.py
dallasfoster: can't do for now.

nvalchemi/distributed/distributed_model.py
dallasfoster: removed.

DD_SCALING_SWEEP_2026-07-01.md
dallasfoster: removed.

presentation/SPEAKER_NOTES.md
dallasfoster: removed.

examples/distributed/06_mace_npt_distributed.py
dallasfoster: done.

benchmark/distributed/fairchem_worker_peakmem.patch
dallasfoster: removed.

dallasfoster:
/ok to test [153a15c](https://github.com/NVIDIA/nvalchemi-toolkit/pull/122/commits/153a15c29ee66a823e42b98ea17620122d48c7ff)

laserkelvin (approved):

dallasfoster:
/ok to test [ff45b3e](https://github.com/NVIDIA/nvalchemi-toolkit/pull/122/commits/ff45b3e5f2e7cbdf98feec646410117524d93bea)

dallasfoster:
/ok to test [78d8e88](https://github.com/NVIDIA/nvalchemi-toolkit/pull/122/commits/78d8e88c90108bea3a9ba1cbe5d2c5c90cd056cc)

dallasfoster:
/ok to test [536c767](https://github.com/NVIDIA/nvalchemi-toolkit/pull/122/commits/536c767d74d6c0b49ab30188ea095e4ad6a38642)

laserkelvin (approved):

PR #127, opened 2026-07-03
Add opt-in hook to exempt fields from the fp-dtype downcast

EricZQu:

## Description

Add opt-in hook to exempt fields from the fp-dtype downcast.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Changes Made

AtomicData.check_fp_dtype_consistency casts every floating-point tensor to positions.dtype (typically float32) so the model computes in one precision. That is the right default for compute inputs, but it also downcasts label fields: e.g. total energy is extensive (~1e4-1e5 eV for large systems), so float32 quantizes it to ~1e-2 eV and reconstructed energies come out discrete/staircased downstream.

Add a _precision_preserving_keys ClassVar and skip those fields in the cast. It is empty by default, so behavior is unchanged; a subclass can opt a high-precision label out of the downcast, e.g.

    class MyData(AtomicData):
        _precision_preserving_keys = frozenset({"energy"})

The exemption holds across add_system_property / add_node_property, since validate_assignment=True re-runs the validator on every setattr.

## Testing

## Checklist

## Additional Notes

> [!TIP]

### DIFF
--- a/nvalchemi/data/atomic_data.py
+++ b/nvalchemi/data/atomic_data.py
@@ -64,6 +64,12 @@ def _warn_first(seen: set, key: object, message: str, *, stacklevel: int = 2) ->
         warnings.warn(message, UserWarning, stacklevel=stacklevel + 1)
 
 
+# Process-global record of (field, src_dtype, tgt_dtype) casts already warned
+# about, so check_fp_dtype_consistency warns once per cast rather than per
+# construction. Module-level so it stays out of the AtomicData schema.
+_FP_CAST_WARNED: set[tuple[str, str, str]] = set()
+
+
 class AtomicNumberTable:
     """
     Atomic number table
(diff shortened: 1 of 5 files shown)


EricZQu:
Added tests for batch, and a first time warning for casting

nvalchemi/data/atomic_data.py
```diff
@@ -366,6 +377,16 @@ class AtomicData(BaseModel, DataMixin):
...
 
+    # FP fields exempt from the positions-dtype cast below. Empty by default (all
+    # fp tensors match positions); a subclass may set it to keep a high-precision
+    # label, e.g. {"energy"} so a fp64 total energy (~1e4-1e5 eV, which fp32 would
+    # quantize to ~1e-2 eV) is not silently downcast to the compute precision.
+    _precision_preserving_keys: ClassVar[frozenset[str]] = frozenset()
+
+    # Process-global (field, src_dtype, tgt_dtype) casts already warned about, so
+    # check_fp_dtype_consistency warns once per cast rather than per construction.
+    _fp_cast_warned: ClassVar[set[tuple[str, str, str]]] = set()
```
laserkelvin: Ooof I'm not a big fan of this since it clutters up the schema...

I think I would be okay to have `_FP_CAST_WARNED` as like a module-level variable that tracks instead of it belonging to the schema (even though it's nominally a `ClassVar`).

laserkelvin (approved):
I'll conditionally approve - I think I agree with the behavior change, but I'm not 100% set on how it's being done here.

I'm not 100% sure if this is best as part of the private API: the way you are using this in the test is by subclassing `AtomicData` which I think is not unreasonable, but I wonder if we should just let users be able to change the precision preservation without jumping through this hoop.

I'm considering these two scenarios:

- You modify the set once per Python session, so that the change in casting behavior is global - you can do this simply by making `_precision_preserving_keys` public
- You are working in a notebook environment, and you want to change the casting behavior temporarily for some subset of the data - for this you would need to have it as instance level

Can you consider the latter case? We can mark it out of scope for now, but I just want you to try and think of a good solution

laserkelvin:
/ok to test [ac8b6d2](https://github.com/NVIDIA/nvalchemi-toolkit/pull/127/commits/ac8b6d2b4adc6fe3ae79c5fa31c6f2430b404d85)

nvalchemi/data/atomic_data.py
EricZQu: Fixed!

EricZQu:
Thanks a lot for the comments @laserkelvin! The API is public now and the warned set is now global. I thought about the per-instance "notebook" use case, and I think maybe we could use a context manager (`with AtomicData.preserving({"energy"}): ...`) for temporarily preserving a field without setting the global flag. It's doable but instance state isn't in model_dump, so a .to() device move drops the exemption. If we want the notebook ergonomics then we need to rewrite `.to()` for `AtomicData`. I could do it but idk if it's worth it.

PR #140, opened 2026-07-18
refactor(dynamics): remove redundant GPUBuffer batch_ptr capacity workaround

### LINKED ISSUE #136: 🐛[BUG]: A pre-allocated Batch.empty buffer can only be filled once: every put() after the first is a silent no-op
### Version

0.1.0 (also present on the latest `main` checkout)

Pip, Source

### Describe the issue

A buffer created with `Batch.empty(...)` accepts exactly one `put()`. After the first successful `put()`, every subsequent `put()` silently copies nothing: `num_graphs` stops growing and no error or warning is raised. `zero()` is not required to trigger this: a plain `put → put` sequence drops the second copy too. The repro below uses `put → zero → put` only because it makes the symptom most visible (`num_graphs` stays `0`).

Expected: a buffer with remaining capacity should accept further `put()`s, and `zero()` should restore the freshly-allocated state. At minimum, a dropped copy should raise or warn rather than silently no-op.

This also affects in-tree consumers that reuse a `Batch.empty` buffer:

- `GPUBuffer` uses the same pattern but is not affected in practice: it already carries an explicit workaround (`_restore_batch_ptr_capacity`) that re-extends the trimmed `batch_ptr` after every write. This workaround was introduced in #4, so the trimming behavior itself seems to be known; fixing it in `put` would make the workaround removable.
- The distributed pipeline: `_CommunicationMixin` reuses pre-allocated send/recv buffers across steps with this exact pattern (`_recv_to_batch` does `put` then `zero`; `_batch_to_buffer` does repeated `put`), so every transfer after the first one would be silently dropped.

```python
import torch
from ase.build import molecule

def atoms_to_data(atoms) -> AtomicData:
    data = AtomicData.from_atoms(atoms)
    data.forces = torch.zeros(data.num_nodes, 3)
    data.energy = torch.zeros(1, 1)
    return data


def put_one(buf: Batch, name: str) -> int:
    src = Batch.from_data_list([atoms_to_data(molecule(name))])
    buf.put(src, mask=torch.ones(1, dtype=torch.bool))
    return buf.num_graphs


template = Batch.from_data_list([atoms_to_data(molecule("H2O"))])
buf = Batch.empty(num_systems=4, num_nodes=64, num_edges=0,
                  template=template, device="cpu")

print("put #1 -> num_graphs =", put_one(buf, "H2O"))

buf.zero()
print("zero() -> num_graphs =", buf.num_graphs)

print("put #2 -> num_graphs =", put_one(buf, "CH4"), " (expected 1)")
print("put #3 -> num_graphs =", put_one(buf, "NH3"), " (expected 2)")
```

### Relevant log output

```shell
put #1 -> num_graphs = 1
zero() -> num_graphs = 0
put #2 -> num_graphs = 0  (expected 1)
put #3 -> num_graphs = 0  (expected 2)
```

### Environment details

```shell
Device-independent: reproduces on both CPU and GPU.
Not tied to any specific CUDA / GPU / PyTorch / Warp version.
```

### Additional context

I have already tracked down the root cause (the first successful `put()` truncates the pre-allocated `batch_ptr` in `SegmentedLevelStorage.put`, so all later capacity checks fail) and have a fix in hand. I understand direct code contributions are not accepted during the public beta. Happy to open a PR once they are, or earlier if a maintainer assigns this issue to me.

lil-lon:

## Description

This PR originally fixed #136. The root cause has since been fixed on `main`
by #141, so this PR is now trimmed down to removing the `GPUBuffer` workaround
made redundant by that fix, plus regression tests. No functional changes.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Related Issues

Relates to #136 (fixed by #141).

## Changes Made

- Remove `GPUBuffer._restore_batch_ptr_capacity()` and its call sites: `put`
  no longer trims `batch_ptr`, so there is nothing to restore.
- Remove the setup-time `_lazy_init_batch_ptr()` trigger loop: `Batch.empty`
  allocates `batch_ptr` eagerly, so it is never `None` for these buffers.
- Add regression tests for the #136 symptom: repeated `put()` into a
  pre-allocated buffer keeps appending.

## Testing

## Checklist

## Additional Notes

CHANGELOG intentionally not updated: no user-visible behavior change.


lil-lon:
#141 fixes the same root cause (#136) by removing the trim in `put()`, so the `level_storage.py` change here becomes redundant once it merges.
Two parts of this PR stay relevant either way:

- the removal of the dead workaround in `sinks.py`
- the regression tests, including repeated `put()` without `zero()`

For the record, why this PR kept the trim: without it, every put leaves the target's `batch_ptr` inconsistent. Its length no longer matches the active segment count, and entries past the last written boundary keep stale values.
No current consumer reads that state, which is why #141 fixes #136.

Restoring capacity instead keeps "trimmed at rest" as a structural guarantee rather than a usage convention. That was the intent.

Happy to close this once #141 lands, or rebase it down to the sinks cleanup + tests if those are still useful.

laserkelvin:
Thank you for the PR @lil-lon - since #141 came up in our own bug catching/fixing efforts, I'd prefer we merge that first, and then ask you to update your PR because I think we are still interested in the cleanup + tests you mentioned.

lil-lon:
Thanks for taking a look!
Since #141 landed on 0.2.0-rc branch, I'll hold off until it is merged to main and then trim this PR down to the cleanup and tests.

lil-lon:
@laserkelvin 
As discussed, now that #141 has landed on main, I've updated this PR.
Could you please take another look when you have time?

laserkelvin:
/ok to test [fc542d1](https://github.com/NVIDIA/nvalchemi-toolkit/pull/140/commits/fc542d1f10fb9eb8996e6e6ea563936b47a7f2b5)

laserkelvin (approved):
Looks good to me - thank you very much for contributing!

PR #149, opened 2026-07-24
feat(training): add LoRA fine-tuning strategy

ys-teh:

## Description

Adds LoRA-based parameter-efficient fine-tuning support within `FineTuningStrategy` for pretrained MLIP models. The implementation supports `peft_config=LoRAConfig(...)`, equivariant LoRA wrappers for e3nn/cuEquivariance layers, trainable-state PEFT checkpoint loading, base-model fingerprint validation, and a MACE LoRA fine-tuning example.

FSDP support is out of scope for this PR. CLI support will be added in a separate PR.

## Type of Change

- [ ] Performance improvement
- [x] Documentation update

## Related Issues

N/A

## Changes Made

- Updated existing training/checkpoint classes to support PEFT fine-tuning:
  - Added `peft_config` support to `FineTuningStrategy`, with LoRA as the currently supported PEFT method.
  - Added trainable and managed parameter registration to `FineTuningStrategy`.
  - Updated module patch behavior so patched modules are trainable by default and do not need to be repeated in `trainable_patterns`.
  - Added `save_trainable_state_only` support to `TrainingStrategy.save_checkpoint` and `CheckpointHook` for parameter-efficient checkpoint saving.
  - Updated checkpoint restore/loading to support trainable-state PEFT checkpoints and partial model state loading.
  - Updated `ModulePatchHook` and `TrainableParameterHook` to use registered trainable/managed parameter names.
  - Added `FineTuningSummaryHook` to provide a concise report on trainable parameters.
  - Added base fingerprint metadata and compatibility checks for loading PEFT checkpoints into base models.
  - Fixed `DemoModelWrapper.from_checkpoint` / `export_model` full-model round-tripping.
- Added LoRA PEFT support for workflows with `LoRAConfig`, fine-tuning serialization, checkpoint validation, PEFT checkpoint loading, and merge-into-base support.
- Added LoRA PEFT support:
  - Added `LoRAConfig` support for configuring LoRA adapter injection through `FineTuningStrategy`.
  - Added merge-into-base support for LoRA workflows.
  - Added reusable LoRA registration hooks and built-in wrapper registration.
  - Added LoRA wrappers for `e3nn.o3.Linear`, e3nn fully connected layers, and cuEquivariance linear layers.
- Added fine-tuning docs and `examples/intermediate/08_lora_finetuning.py` for MACE `medium-mpa-0` LoRA fine-tuning on LPSC data.

## Testing

Targeted tests run:

- `uv run --extra cu12 pytest test/training/test_peft_lora.py test/training/test_finetune.py test/training/test_reference_energies.py test/training/test_strategy.py test/training/test_checkpoint.py
` (passed: 252 tests)

## Checklist

- [ ] I have performed a self-review of my code

## Additional Notes

Benchmarking against published results:

image

Simple benchmarking sanity check with `TRAINING_EPOCHS=100` using the provided LoRA example, compared against a similar setup trained from scratch:

image

This currently depends on a PhysicsNeMo version with the PEFT LoRA support used here. The version pin can be revisited and updated as the required PEFT APIs become available in a newer stable PhysicsNeMo release.

> [!TIP]


nvalchemi/training/_checkpoint.py
```diff
@@ -1183,7 +1264,12 @@ def _restore_checkpoint_into_strategy(
...
-        model.load_state_dict(weights)
+        _load_model_state(
+            model,
+            weights,
+            model_name=name,
+            strategy_metadata=strategy_metadata,
+        )
         spec_path = root / "models" / name / "spec.json"
         spec = _load_spec(spec_path) if spec_path.exists() else None
         loaded_models[name] = (model, spec)
```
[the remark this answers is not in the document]
ys-teh: Added comment on sanity check for clarity ([dc2d1d3](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/dc2d1d3ee2e31782d48be426cac1bc4e03e29d15)).

nvalchemi/training/peft/lora.py
```diff
@@ -0,0 +1,1062 @@
...
+                f"{model_name!r}: {sorted(load_result.unexpected_keys)}."
+            )
+
+        # Check if all the saved keys were loaded.
+        missing_saved_keys = set(combined_state) & set(load_result.missing_keys)
+        if missing_saved_keys:
+            raise RuntimeError(
+                "LoRA adapter saved tensor keys were not loaded for model "
+                f"{model_name!r}: {sorted(missing_saved_keys)}."
+            )
```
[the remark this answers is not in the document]
ys-teh: Added comment on sanity check for clarity ([dc2d1d3](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/dc2d1d3ee2e31782d48be426cac1bc4e03e29d15)). And created `load_partial_model_state` for reuse.

nvalchemi/training/peft/lora.py
```diff
@@ -0,0 +1,1062 @@
...
+        registered_patch_parameter_names = set(
+            registered_trainable_names.get(_MODULE_PATCH_HOOK_IDENTIFIER, ())
+        )
+        extra_parameter_names = set(self._optimizer_parameter_names) - (
+            registered_parameter_names
+        )
+        patch_targets = tuple(self.module_patches)
+
+        manifest_models: dict[str, dict[str, Any]] = {}
+        for model_name in model_names:
```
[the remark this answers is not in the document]
ys-teh: Updated in [dc2d1d3](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/dc2d1d3ee2e31782d48be426cac1bc4e03e29d15).

examples/intermediate/08_lora_finetuning.py
```diff
@@ -0,0 +1,704 @@
...
+# Similar to :class:`~nvalchemi.training.TrainingStrategy` and
+# :class:`~nvalchemi.training.FineTuningStrategy`,
+# :class:`~nvalchemi.training.LoRAFineTuningStrategy` combines the training
+# objective, validation reporter, neighbor-list hook, and optimizer setup. The
+# LoRA-specific target patterns select the MACE layers that receive adapters,
+# while the trainable patterns keep a small set of base-model parameters
+# trainable. Module patches can also be provided to customize model components,
+# but they are not needed in this example.
+
+seed_everything(SEED)
```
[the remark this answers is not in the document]
ys-teh: Made `is_lora_layer` public in [7e61113](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/7e61113fbcd765a87f2a2115a6b227c1bb1e894d).

nvalchemi/training/peft/wrappers.py
```diff
@@ -0,0 +1,783 @@
...
+            )
+            continue
+        _peft.register_lora_wrapper(layer_cls, wrapper_cls)
+    _BUILTIN_LORA_WRAPPERS_REGISTERED = True
+
+
+def available_lora_wrappers() -> LoRAWrapperRegistrations:
+    """Return registered (layer, LoRA-wrapper) pairs."""
+    register_builtin_lora_wrappers()
+    return tuple(_peft._physicsnemo_peft.lora._LORA_WRAPPERS.items())
```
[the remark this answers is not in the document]
ys-teh: Moved import in [0b487f2](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/0b487f2a7cb2a3f9a6d0dd7a608e11872c439399).

nvalchemi/training/peft/lora.py
```diff
@@ -0,0 +1,1048 @@
...
+                    f"{current_fingerprint!r}."
+                )
+                if strict:
+                    raise ValueError(message)
+                warnings.warn(message, UserWarning, stacklevel=2)
+
+        # Register LoRA wrappers and apply LoRA adapters according to the saved strategy.
+        register_builtin_lora_wrappers()
+        for layer_cls, wrapper_cls in cls._wrapper_registrations_from_spec(strategy):
+            _peft.register_lora_wrapper(layer_cls, wrapper_cls)
```
zubatyuk: A trust policy shoud be enforced for the custom classes in strategy.json. Coincidently, there is a related PR in another project: https://github.com/isayevlab/aimnetcentral/pull/108

nvalchemi/training/peft/lora.py
```diff
@@ -0,0 +1,1048 @@
...
+    if not isinstance(metadata, Mapping):
+        return
+    saved = metadata.get("base_model_fingerprints")
+    if not isinstance(saved, Mapping):
+        return
+    normalized = strategy_validation._normalize_models(models)
+    if not isinstance(normalized, Mapping):
+        return
+    current = {
+        name: _peft.compute_base_fingerprint(model)
```
zubatyuk: Note, this computes hash of the architecture, not the weights. LoRA weights should be tied to the weights of the model.

nvalchemi/training/peft/lora.py
```diff
@@ -0,0 +1,1048 @@
...
+        extra_parameter_names = set(self._optimizer_parameter_names) - (
+            registered_parameter_names
+        )
+        patch_targets = tuple(self.module_patches)
+
+        manifest_models: dict[str, dict[str, Any]] = {}
+        for name in model_names:
+            # Collect this model's patch targets, metadata, and trainable states.
+            patch_prefix = f"{name}."
+            named_parameters = dict(checkpoint_models[name].named_parameters())
```
zubatyuk: Not commonly used in MLIPs, but buffers like `running_mean` and `running_var` for `BatchNorm` should also be exported.

nvalchemi/training/peft/lora.py
zubatyuk: https://github.com/zubatyuk/aimnetcentral/blob/feat/artifact-trust-boundary/docs/model_format.md#model-yaml-import-policy

nvalchemi/training/peft/lora.py
ys-teh: Thanks for pointing it out. I have now added buffers in `nvalchemi.training._checkpoint._filter_snapshot_to_trainable_state`.

nvalchemi/training/peft/lora.py
ys-teh: Good suggestion! In the refactored version, I have now introduced `allowed_import_paths` and `trust_remote_code` controls in `nvalchemi/training/peft/loading.py`, specifically for `load_peft_checkpoint_into_model`.

nvalchemi/training/peft/lora.py
ys-teh: Yes, I am aware of that. I have not added weight hashes yet as I am still thinking through the best way to do it without hurting scalability for large models.

nvalchemi/training/peft/__init__.py
dallasfoster, line 1: The lazy `__getattr__` blocks in this file, `training/__init__.py`,
`hooks/__init__.py` and `lora_wrappers.py` are all defeated by one eager import.
Every hop below is module-level:

```
nvalchemi/training/__init__.py          → finetune
nvalchemi/training/finetune.py:41       → peft.fingerprints
nvalchemi/training/peft/__init__.py:19  → peft.lora_wrappers      ← this line
nvalchemi/training/peft/lora_wrappers.py:31 → peft._peft
nvalchemi/training/peft/_peft.py:19     → physicsnemo.experimental.peft
```

A user importing `nvalchemi.training` for plain from-scratch training, who has
never heard of PEFT, currently gets:

```python
import sys, warnings
with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")
    from nvalchemi.training import TrainingStrategy
# ExperimentalFeatureWarning: You are importing from 'physicsnemo.experimental'.
# The APIs in this namespace are experimental, under active development, and may
# change without notice.
```

nvalchemi/training/finetune.py
```diff
@@ -228,25 +291,203 @@ def _prepend_finetuning_hooks(cls, data: Any) -> Any:
...
+                    f"got {type(peft_config).__name__}."
+                )
+            if tuple(normalized.get("freeze_patterns") or ()):
+                raise ValueError(
+                    "FineTuningStrategy with peft_config does not accept "
+                    "freeze_patterns; PEFT freezes the base model by default."
+                )
+            if normalized.get("compute_base_fingerprints", True):
+                generated.append(BaseFingerprintHook())
+            generated.extend(peft_setup_hooks(peft_config, normalized))
```
dallasfoster, line 313: `_validate_generated_hooks` intercepts any dict on `peft_config` and routes it
through `peft_config_from_metadata`, which requires an envelope whose inner key
repeats the field name. So the natural Pydantic idiom fails — and so does the
config's own serialized form:

```python
# settings loaded from a JSON config file
lora_settings = {"rank": 8, "alpha": 1.0, "lora_target_patterns": ["main.model.projection"]}
FineTuningStrategy(**base, peft_config=lora_settings)
# ValidationError: Value error, Unsupported PEFT method None.

# adding the discriminator isn't enough either
lora_settings["peft_method"] = "lora"
FineTuningStrategy(**base, peft_config=lora_settings)
# ValidationError: Value error, PEFT metadata must contain a peft_config object.

# only the doubly-nested envelope is accepted
peft_config={"peft_method": "lora", "peft_config": {...}, "peft_config_class": ...}
```

Suggestion: make the field a Pydantic v2 discriminated union on `peft_method`
and keep the envelope strictly inside `to_spec_dict`/`from_spec_dict`.

nvalchemi/training/peft/config.py
```diff
@@ -0,0 +1,135 @@
...
+    -----
+    PEFT metadata uses ``peft_method`` as the loading discriminator. Any
+    recorded config class path is provenance metadata only and must not be
+    dynamically imported during load.
+    """
+    from nvalchemi.training.peft.lora import LORA_PEFT_METHOD
+
+    if config.peft_method == LORA_PEFT_METHOD:
+        from nvalchemi.training.peft.lora import lora_metadata_from_config
+        return lora_metadata_from_config(config, details)
```
dallasfoster, line 80: Four sites each doing `if method == LORA_PEFT_METHOD: ... else: raise
ValueError("Unsupported PEFT method")`:

Adding DoRA or IA³ means finding and editing all four, with nothing structural
to make you. The toolkit already solved this twice — `register_type_serializer`,
and `BaseSpec`/`create_model_spec_from_json`.

nvalchemi/training/finetune.py
```diff
@@ -228,25 +291,203 @@ def _prepend_finetuning_hooks(cls, data: Any) -> Any:
...
+            if not isinstance(peft_config, PeftConfig):
+                raise TypeError(
+                    "FineTuningStrategy peft_config must be a PeftConfig; "
+                    f"got {type(peft_config).__name__}."
+                )
+            if tuple(normalized.get("freeze_patterns") or ()):
+                raise ValueError(
+                    "FineTuningStrategy with peft_config does not accept "
+                    "freeze_patterns; PEFT freezes the base model by default."
+                )
```
dallasfoster, line 310: - `freeze_patterns` **must be empty** when `peft_config` is set (raises here).
- `compute_base_fingerprints` is documented "Used only when `peft_config` is
  provided" — inert otherwise.
- `trainable_patterns` changes meaning ("With `peft_config`, use only for extra
  parameters beyond adapters and module patches").

The toolkit's usual answer to this shape is a nested config object —
`ValidationConfig`, `OptimizerConfig`. `compute_base_fingerprints` is a property
of the PEFT run and would sit naturally on `PeftConfig`; moving it also removes
the collision between the bool field `FineTuningStrategy.compute_base_fingerprints`
and the function `nvalchemi.training.peft.compute_base_fingerprints(models)`,
which currently differ only by context.

nvalchemi/training/peft/lora_hook.py
```diff
@@ -0,0 +1,264 @@
...
+            workflow, "_flat_opts", None
+        ):
+            raise RuntimeError(
+                "LoRAHook must be registered before optimizers are built."
+            )
+
+        # Register built-in and user-defined LoRA wrappers
+        register_builtin_lora_wrappers()
+        for layer_cls, wrapper_cls in self.lora_config.wrapper_registrations or ():
+            _peft.register_lora_wrapper(layer_cls, wrapper_cls)
```
dallasfoster, line 160: `register_lora_wrapper` mutates a plain module-level dict inside PhysicsNeMo
(`_peft.py:36` → `physicsnemo.experimental.peft.lora._LORA_WRAPPERS`), and
`LoRAConfig.wrapper_registrations` are installed into it at hook-registration
time — i.e. as a side effect of constructing a strategy. Two runs in one process
share one registry:

```python
run_a = finetune(adapter=CheapAdapter)   # registers MyLayer -> CheapAdapter
run_b = finetune(adapter=RichAdapter)    # registers MyLayer -> RichAdapter

run_a.peft_config.wrapper_registrations[0][1].__name__   # 'CheapAdapter'
[w.__name__ for l, w in available_lora_wrappers() if l is MyLayer]   # ['RichAdapter']
```

`run_a`'s config is unchanged and still names `CheapAdapter`, but re-running it —
a notebook cell, a sweep loop, a retry — now silently builds `RichAdapter`. No
warning, and there's no unregister API to undo it.

Compounding this: `register_builtin_lora_wrappers()` is latched by a module
global (`lora_wrappers.py:71`), so if anything perturbs the registry afterwards
a second call is a no-op and won't restore the builtins.

nvalchemi/training/peft/fingerprints.py
```diff
@@ -0,0 +1,93 @@
...
+        return
+
+    def on_register(self, workflow: Any) -> None:
+        """Store base model fingerprints on ``workflow``."""
+        models = getattr(workflow, "models", None)
+        if not isinstance(models, Mapping):
+            raise TypeError(
+                "BaseFingerprintHook requires a workflow with a models mapping."
+            )
+        workflow._base_fingerprints = compute_base_fingerprints(models)
```
dallasfoster, line 93: `LoRAHook` registers through the public `register_*_parameter_names` methods and
raises a clear `TypeError` when they're missing (`lora_hook.py:247-256`) — good.
But `BaseFingerprintHook` here and `FineTuningSummaryHook`
(`hooks/finetune.py:396`) assign private attributes on the workflow directly:

```python
workflow._base_fingerprints = compute_base_fingerprints(models)
workflow._trainable_parameter_summary = _trainable_parameter_summary(workflow, models)
```

Same PR, same hook family, two contracts. I'd standardize on the public-method
one.

nvalchemi/training/hooks/__init__.py
```diff
@@ -29,12 +33,28 @@
...
     "TorchProfilerHook",
+    "FineTuningSummaryHook",
     "ModulePatchHook",
     "TrainableParameterHook",
     "TrainingUpdateHook",
     "TrainingUpdateOrchestrator",
 ]
+
+
+def __getattr__(name: str) -> object:
```
laserkelvin: What's the reason for needing to do this lazily?

nvalchemi/training/peft/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    if name == "load_peft_checkpoint_into_model":
+        from nvalchemi.training.peft import loading
+
+        return getattr(loading, name)
+    if name == "is_lora_layer":
+        from nvalchemi.training.peft import _peft
+
+        return _peft.is_lora_layer
+    if name in {
+        "CuEquivariantLoRALinear",
```
laserkelvin: The "Equivariant" part to me is a little counterintuitive

either `CuEqLoRALinear` for short hand, or `CuEquivarianceLoRALinear` to match what the library is called

nvalchemi/training/hooks/ema.py
```diff
@@ -222,6 +222,8 @@ class EMAHook(BaseModel, TrainingUpdateHook):
 
     _averaged_model: AveragedModel | None = PrivateAttr(default=None)
     _pending_averaged_state: dict[str, Any] | None = PrivateAttr(default=None)
+    # Preserve strict-vs-partial load intent until lazy EMA initialization runs.
+    _pending_averaged_state_load: str = PrivateAttr(default="full")
```
laserkelvin: Instead of `str` as the hint, I'd prefer `Literal["full", "partial"]`

Even if `PrivateAttr` doesn't validate against the values, it's good to document like this

nvalchemi/training/hooks/ema.py
```diff
@@ -379,13 +387,20 @@ def load_state_dict(self, state: Mapping[str, Any]) -> None:
         if "num_updates" in state:
             self.num_updates = int(state["num_updates"])
         if "averaged_model_state" in state:
+            averaged_state_load = str(state.get("averaged_model_state_load", "full"))
```
laserkelvin: Why does it need to cast to `str`?

nvalchemi/training/__init__.py
```diff
@@ -66,6 +70,7 @@
     step_optimizers,
     zero_gradients,
 )
+from nvalchemi.training.reference_energies import fit_atomic_reference_energies
```
laserkelvin: Out of scope for this PR, but I really think we should re-design this into a workflow better. Arguably, the atomic energy reference part should be part of an ETL pipeline like with `physicsnemo-curator`, but I respect that we should have some kind of interface to that within Toolkit.

I think the main thing is letting users have the ability to customize how and what they extract, knowing that this is a fairly specific thing being done for MACE right now

nvalchemi/training/__init__.py
```diff
@@ -108,19 +115,40 @@
...
     "default_training_fn",
+    "fit_atomic_reference_energies",
     "freeze_unconfigured_models",
+    "is_lora_layer",
+    "load_peft_checkpoint_into_model",
     "loss_component_to_spec",
     "load_checkpoint",
     "move_to_devices",
     "register_type_serializer",
+    "register_builtin_lora_wrappers",
```
laserkelvin: Do the `peft` specific things need to be exported from `nvalchemi.training`? Can't they just be available from `nvalchemi.training.peft`?

I worry about crowding this namespace too much

nvalchemi/training/_checkpoint.py
```diff
@@ -438,6 +438,123 @@ def _snapshot_components(
     }
 
 
+def _selected_state_dict(
+    state_dict: dict[str, Any],
```
laserkelvin: Is this actually `Any`, or should they all be `tensor`?

nvalchemi/training/_checkpoint.py
```diff
@@ -438,6 +438,123 @@ def _snapshot_components(
...
+    names: set[str],
+) -> dict[str, Any]:
+    """Return selected state entries and fail when metadata names are stale."""
+    missing = sorted(names - set(state_dict))
+    if missing:
+        raise KeyError(
+            f"Cannot checkpoint missing trainable parameter(s): {missing!r}."
+        )
+    return {name: state_dict[name] for name in sorted(names)}
```
laserkelvin: You could just do `sorted(state_dict.items())` directly

nvalchemi/training/_checkpoint.py
```diff
@@ -438,6 +438,123 @@ def _snapshot_components(
...
+
+
+def _model_local_parameter_names(model_name: str, names: Iterable[str]) -> list[str]:
+    """Return names local to ``model_name`` from fully qualified parameters."""
+    prefix = f"{model_name}."
+    return [name.removeprefix(prefix) for name in names if name.startswith(prefix)]
```
laserkelvin: I don't think you need `.startswith` if you are using `.removeprefix` - it'll be no-op if it doesn't start

nvalchemi/training/_checkpoint.py
```diff
@@ -438,6 +438,123 @@ def _snapshot_components(
...
+    """Return selected state entries and fail when metadata names are stale."""
+    missing = sorted(names - set(state_dict))
+    if missing:
+        raise KeyError(
+            f"Cannot checkpoint missing trainable parameter(s): {missing!r}."
+        )
+    return {name: state_dict[name] for name in sorted(names)}
+
+
+def _model_local_parameter_names(model_name: str, names: Iterable[str]) -> list[str]:
```
laserkelvin: This method is only called once by `_filter_snapshot_to_trainable_state` - this is a pattern by GPT models I'm not a fan of, and would rather we just move this logic into that caller

In general, could you try and catch it when this happens?

laserkelvin (commented):
Initial wave of comments, more to come in a bit

examples/intermediate/08_lora_finetuning.py
```diff
@@ -0,0 +1,715 @@
...
+# validation splits for reproducibility.
+
+if len(samples) < 2:
+    raise ValueError("Need at least two structures for train/validation split.")
+generator = torch.Generator().manual_seed(SEED)
+shuffled = torch.randperm(len(samples), generator=generator).tolist()
+validation_count = max(1, int(round(len(samples) * VALIDATION_FRACTION)))
+validation_indices = sorted(shuffled[:validation_count])
+train_indices = sorted(shuffled[validation_count:])
+train_batch = Batch.from_data_list([samples[index] for index in train_indices])
```
laserkelvin: This is a really minor comment, but you could actually just construct one big `Batch`, and then create `train_batch` and `validation_batch` by slicing, instead of looping. Consider it refactoring to show users better performance

examples/intermediate/08_lora_finetuning.py
```diff
@@ -0,0 +1,715 @@
...
+def make_loader(
+    batch: Batch,
+    *,
+    batch_size: int,
+    shuffle: bool,
+) -> DataLoader:
+    """Create a DataLoader over an in-memory batch."""
+    dataset = InMemoryDataset(
+        in_memory_batch=batch,
+        device=device,
```
laserkelvin: This is taking `device` from out of function scope

examples/intermediate/08_lora_finetuning.py
```diff
@@ -0,0 +1,715 @@
...
+    )
+    return DataLoader(
+        dataset,
+        batch_size=batch_size,
+        shuffle=shuffle,
+        drop_last=shuffle,
+        prefetch_factor=0,
```
laserkelvin, line 257: Interesting that these need to be specified for an `InMemoryDataset` - is this a problem with our `DataLoader` implementation then?

nvalchemi/training/hooks/__init__.py
ys-teh: It's mainly because the peft implementation in physicsnemo is still considered experimental, and partially because I'm currently relying on a specific git version to be able to import peft from physicsnemo. However, given that physicsnemo will be released earlier than toolkit, I can get rid of the lazy import and merge this PR after the upcoming physicsnemo release. Let me know!

examples/intermediate/08_lora_finetuning.py
```diff
@@ -0,0 +1,715 @@
...
+    trainable_patterns=TRAINABLE_PATTERNS,
+)
+
+
+# %%
+# After the strategy attaches adapters, inspect the inserted LoRA modules before
+# training.
+
+
+def print_lora_adapters(strategy: FineTuningStrategy) -> None:
```
laserkelvin: Why does this need to be its own function?

nvalchemi/training/peft/__init__.py
ys-teh: Good catch! Updated in [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457). Also added `test_finetuning_import_does_not_load_experimental_peft` in `test_finetune.py`.

nvalchemi/training/peft/__init__.py
ys-teh: Updated in [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457).

nvalchemi/training/hooks/ema.py
ys-teh: Updated in [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457).

nvalchemi/training/hooks/ema.py
ys-teh: Updated in [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457).

nvalchemi/training/__init__.py
ys-teh: Not really, moved all peft-related items to `nvalchemi.training.peft`. [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457)

nvalchemi/training/_checkpoint.py
ys-teh: Good point, it should be tensor. I also deleted this function entirely so it stays inline in `_filter_snapshot_to_trainable_state`. [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457)

nvalchemi/training/_checkpoint.py
ys-teh: In this case, we are returning a subset of `state_dict` that are previously trained. So I have to keep the previous behavior but renamed `names` to `model_trainable_names` for better clarity. [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457)

nvalchemi/training/_checkpoint.py
ys-teh: Here we need to filter based on the prefix (or model name) and that is why the `.startswith` check is there. I have also moved this function inline in `_filter_snapshot_to_trainable_state`. [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457)

nvalchemi/training/_checkpoint.py
ys-teh: Yes, I also notice that pattern and don't like how it makes things less readable. Thanks for pointing this out. I have made the other 2 functions inline in `_fliter_snapshot_to_trainable_state` as well.

nvalchemi/training/_checkpoint.py
laserkelvin: No but my point is that, you don't necessarily need to have the `if` check to filter, because running `removeprefix` directly will be no-op if there isn't a prefix

nvalchemi/training/peft/lora_hook.py
ys-teh: Good point. I have modified the wrapper registration behavior in [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457). Now the registration stays temporary in the lora hook so it doesn't affect the global behavior. And a new lora wrapper can now override an old one but with a warning. See the updated `_register_lora_wrappers` in `nvalchemi/training/peft/lora_wrappers.py`.

nvalchemi/training/peft/fingerprints.py
ys-teh: Updated to have the similar register* methods along with errors when items are missing. [890bdec](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/890bdec2cf4a171be4a0ceb947f69eb84bf96457)

nvalchemi/training/__init__.py
ys-teh: I added this helper mainly to keep `examples/intermediate/08_lora_finetuning.py` (including the sphinx version) concise. I realize that models like PET-MAD also need the same step for fine-tuning, so the helper only performs a generic least-squares fit. The MACE specific way of computing delta and correcting afterwards is only shown in the example itself. Where would you recommend to put function like `fit_atomic_reference_energies` in this PR?

laserkelvin: Maybe add a TODO - this feels messy enough that we should try and make sure it doesn't get into a public release (not criticizing your work, just saying that the pattern itself is gnarly)

examples/intermediate/08_lora_finetuning.py
ys-teh: I remember the general guideline is to use InMemoryDataset when the dataset is small so have used that. I have not checked the other case. Is there another preferred way?

laserkelvin: I don't know if there's an alternative way, but I think ideally if they are using an `InMemoryDataset` we don't have to specify kwargs in that particular way. Probably out of scope for this PR, but would be good if we can think about how to make this a better experience

nvalchemi/training/_checkpoint.py
ys-teh: Sorry, I didn't explain clearly earlier. Let's look at one example:
`trainable_names=[modelA.model.tensor1, modelB.model.tensor2]`
Without `.startswith('modelA')`, we will end with
`model_trainable_names=[model.tensor1, modelB.model.tensor2]`, but we need `model_trainable_names=[model.tensor1]` for that specific model.

examples/intermediate/08_lora_finetuning.py
ys-teh: > I don't know if there's an alternative way, but I think ideally if they are using an `InMemoryDataset` we don't have to specify kwargs in that particular way. Probably out of scope for this PR, but would be good if we can think about how to make this a better experience

I think I understand a little better what you were asking. For InMemoryDataset, there is still CPU to GPU transfer for each batch. The default prefetch_factor=2 is not necessary but should work fine. I can remove prefetch_factor=0 to avoid confusion.

laserkelvin: If the dataset is small, maybe we can just move the whole thing into GPU memory, then you won't need to deal with these semantics

zubatyuk: Bad idea. GPU memory should be for the model.

nvalchemi/training/finetune.py
ys-teh: Updated in [9c51ed6](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/9c51ed64eba066ff0b850abe9894d3d0a8b302c4). Now it accepts the second case you showed above, which is also consistent with the updated to_spec_dict result.

nvalchemi/training/peft/config.py
ys-teh: Good catch. I have now added a `register_peft_method`. Its arguments specify exactly all the methods required if one were to implement a new peft method. I have also updated the docs to include the details. [9c51ed6](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/9c51ed64eba066ff0b850abe9894d3d0a8b302c4)

nvalchemi/training/__init__.py
ys-teh: Got it, removed the public import and added TODO in the original file itself. [a3ae46a](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/a3ae46a1ea22df5807f0ce2e60cd797c38e9007f)

examples/intermediate/08_lora_finetuning.py
ys-teh: Sounds good, updated in [a3ae46a](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/a3ae46a1ea22df5807f0ce2e60cd797c38e9007f).

examples/intermediate/08_lora_finetuning.py
ys-teh: Good catch, updated in [a3ae46a](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/a3ae46a1ea22df5807f0ce2e60cd797c38e9007f).

nvalchemi/training/finetune.py
ys-teh: `peft_config` currently contains only configs related to adapter construction, and I'm hoping to keep it focused for PEFT-specific hooks. We could also compute fingerprints for regular fine-tuning, but that is likely unnecessary since the full model weights are typically saved. That said, if it still makes more sense to move `compute_base_fingerprints` to `peft_config`, I’m happy to do that. Let me know.

nvalchemi/training/hooks/__init__.py
laserkelvin: You've addressed my comment but I'll leave this open so we don't forget

laserkelvin (approved):
I'm approving from my end - LGTM!

ys-teh:
/ok to test [c02c40d](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/c02c40d2aeec8bf131c03d292e951bcdc9b1c2be)

ys-teh:
/ok to test [425126a](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/425126ad886698a238fce408fbda014be80f8c56)

laserkelvin:
/ok to test [425126a](https://github.com/NVIDIA/nvalchemi-toolkit/pull/149/commits/425126ad886698a238fce408fbda014be80f8c56)

laserkelvin:
Merge queue version of the multiGPU fails, but working on the PR CI. In either case, fix would be out of scope for this PR, so I think it's safe to merge.

PR #160, opened 2026-08-06
Omnibus Domain Decomposition Bug Fixes

dallasfoster:

## Description

Correctness work on the distributed (domain-decomposition) layer, covering a
reported set of DD defects plus several found while building gates for them.
The common thread is silence: each of these produced wrong numbers or dropped
work without raising, so much of the diff is the gate that now catches it.

Domain decomposition ships for the first time in this release, so most of these
are fixes to unreleased code. Two exceptions change already-released
single-process behaviour and are called out separately below.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Changes Made

Eight commits, each independently reviewable:

1. **Keep the spec whole across a process boundary.** `MLIPSpec.to_dict` omitted
   `CompilePolicy`, so a spec persisted to disk or sent to a worker came back
   with `force_strategy` unset and framework-autograd models differentiated an
   already-reduced global energy — forces scaled by the world size. Adapter
   replacements built from a model's own submodules are closures with no
   importable name, so a reloaded adapter installed as a no-op and per-layer
   halo correction stopped running.
2. **Correct the all-reduce adjoint for replicated consumers.** `AllReduceSum`
   is self-adjoint only when the incoming gradient is this rank's own partial;
   where the reduced value feeds a computation every rank runs identically, the
   gradient is already global and reducing again multiplied it by the world size.
3. **Sharding and halo geometry.** A ghost shell wider than one domain reaches
   ranks more than one domain away, and reaches the same rank from *both* sides
   of a periodic axis — each side a distinct lattice image holding different
   atoms. Only the outermost domain pair was treated as wrapping, so in-cutoff
   pairs were dropped silently. Also: ranks without the source batch chose a
   device from hardware availability rather than the mesh, so a CPU mesh on a
   GPU host mixed CPU and CUDA tensors in its first broadcast and hung.
4. **Consolidate the strain trick; stress and output parity.** The affine strain
   trick had five copies across the distributed layer beside the canonical
   `prepare_strain`; all now route through it and a new `apply_strain`. Stress
   under node-partition graph parallelism previously raised `KeyError`. The
   framework-owned energy returned `[n_systems]` where wrappers return
   `[n_systems, 1]`, and kept the wider dtype node energies accumulate in.
5. **fp32 precision control.** `nvalchemi.distributed.pin_fp32`, plus a warning
   when reduced-precision fp32 is in force — it makes distributed and
   single-process results diverge well beyond fp32 rounding.
6. **Electrostatics gradients and DD-pure specs.** Ewald's reciprocal was only
   differentiated w.r.t. positions and charges, giving a wrong `dE/dq` and a
   wrong strain-autograd stress; the strain cache reused the previous call's
   graph. Ewald/PME default to `hybrid_forces=False` (**breaking**, see below).
7. **Per-spawn rendezvous ports, docs and changelog.** The spawn harnesses
   hardcoded a port, so concurrent runs of the same file collided with
   `EADDRINUSE` and read as correctness failures.
8. **A distinct rendezvous port per spawn test.** Seven ports were shared across
   fifteen files, and the per-parametrization helper drew from the ephemeral
   range.

### Breaking

`EwaldModelWrapper` and `PMEModelWrapper` now default to `hybrid_forces=False`.
The analytic direct-output path does not produce consistent gradients and is
rejected under domain decomposition. Forces and stress now come from autograd
over the energy; pass `hybrid_forces=True` explicitly to keep the old path.

## Testing

Local (1 GPU): **1313 passed, 64 skipped, 0 failed** across `test/distributed`
and `test/models`.

Multi-GPU (4×H100, world=2 and world=4):

| gate | result |
|---|---|
| equivalence matrices, eager + compiled | **32/32 PASS** |
| stress gates (electrostatics, slab ×2, MACE cueq, wired pipeline) | **11 passed** |
| model + graph-parallel gates | **35 passed** |
| full `test/distributed/` | 656 passed, 7 skipped, 3 failed |

The three were rendezvous-port collisions, not correctness failures: two test
files shared a hardcoded port, and the per-parametrization port helper drew from
the kernel's ephemeral range, so a port could be claimed for an unrelated
connection between the choice and the bind. One of the three had passed minutes
earlier in the same job. Both causes are fixed in this branch (every spawn test
now has a distinct port, below the ephemeral floor); the full-suite figure above
predates that fix.

### Force / stress equivalence vs single-GPU — eager (world=2, H100, true fp32)

| model | system | want | N | owned | relE | relF | relS | verdict |
|---|---|---|---|---|---|---|---|---|
| ewald | neutral | F | 2744 | [1352, 1392] | 0.00e+00 | 4.30e-06 | n/a | PASS |
| ewald | neutral | F+S | 2744 | [1352, 1392] | 0.00e+00 | 4.44e-06 | 2.64e-07 | PASS |
| ewald | net-charged | F | 2744 | [1352, 1392] | 0.00e+00 | 4.55e-06 | n/a | PASS |
| ewald | net-charged | F+S | 2744 | [1352, 1392] | 0.00e+00 | 4.26e-06 | 6.59e-07 | PASS |
| pme | neutral | F | 2744 | [1352, 1392] | 0.00e+00 | 5.01e-06 | n/a | PASS |
| pme | neutral | F+S | 2744 | [1352, 1392] | 0.00e+00 | 4.89e-06 | 9.89e-07 | PASS |
| pme | net-charged | F | 2744 | [1352, 1392] | 0.00e+00 | 4.55e-06 | n/a | PASS |
| pme | net-charged | F+S | 2744 | [1352, 1392] | 0.00e+00 | 4.60e-06 | 1.12e-06 | PASS |
| aimnet2+pme | neutral | F | 2560 | [1286, 1274] | 1.66e-06 | 7.31e-07 | n/a | PASS |
| aimnet2+pme | neutral | F+S | 2560 | [1286, 1274] | 1.44e-06 | 4.50e-07 | 2.19e-05 | PASS |
| aimnet2+pme | net-charged | F | 2560 | [1286, 1274] | 2.21e-07 | 2.77e-07 | n/a | PASS |
| aimnet2+pme | net-charged | F+S | 2560 | [1286, 1274] | 3.32e-07 | 4.43e-07 | 2.64e-05 | PASS |
| aimnet2+ewald | neutral | F | 2560 | [1286, 1274] | 1.66e-06 | 3.38e-07 | n/a | PASS |
| aimnet2+ewald | neutral | F+S | 2560 | [1286, 1274] | 1.33e-06 | 2.81e-07 | 2.63e-05 | PASS |
| aimnet2+ewald | net-charged | F | 2560 | [1286, 1274] | 0.00e+00 | 3.60e-07 | n/a | PASS |
| aimnet2+ewald | net-charged | F+S | 2560 | [1286, 1274] | 4.43e-07 | 4.99e-07 | 5.13e-05 | PASS |

### Same, compiled (`graphs` is compile count before → after; stable means recompile-free)

| model | system | want | N | owned | relE | relF | relS | graphs | verdict |
|---|---|---|---|---|---|---|---|---|---|
| ewald | neutral | F | 2744 | 1352 | 0.00e+00 | 4.43e-06 | n/a | 18→18 | PASS |
| ewald | neutral | F+S | 2744 | 1352 | 0.00e+00 | 4.38e-06 | 2.64e-07 | 37→37 | PASS |
| ewald | net-charged | F | 2744 | 1352 | 0.00e+00 | 4.45e-06 | n/a | 37→37 | PASS |
| ewald | net-charged | F+S | 2744 | 1352 | 0.00e+00 | 4.39e-06 | 3.30e-07 | 37→37 | PASS |
| pme | neutral | F | 2744 | 1352 | 0.00e+00 | 4.50e-06 | n/a | 59→59 | PASS |
| pme | neutral | F+S | 2744 | 1352 | 0.00e+00 | 4.56e-06 | 4.62e-07 | 84→84 | PASS |
| pme | net-charged | F | 2744 | 1352 | 0.00e+00 | 4.52e-06 | n/a | 84→84 | PASS |
| pme | net-charged | F+S | 2744 | 1352 | 0.00e+00 | 4.36e-06 | 1.12e-06 | 84→84 | PASS |
| aimnet2+pme | neutral | F | 2560 | 1286 | 1.44e-06 | 3.94e-07 | n/a | 152→152 | PASS |
| aimnet2+pme | neutral | F+S | 2560 | 1286 | 7.75e-07 | 3.38e-07 | 1.64e-05 | 235→235 | PASS |
| aimnet2+pme | net-charged | F | 2560 | 1286 | 1.11e-07 | 3.88e-07 | n/a | 253→253 | PASS |
| aimnet2+pme | net-charged | F+S | 2560 | 1286 | 2.21e-07 | 2.77e-07 | 1.36e-05 | 253→253 | PASS |
| aimnet2+ewald | neutral | F | 2560 | 1286 | 1.77e-06 | 4.50e-07 | n/a | 259→259 | PASS |
| aimnet2+ewald | neutral | F+S | 2560 | 1286 | 1.11e-06 | 4.50e-07 | 2.32e-05 | 264→264 | PASS |
| aimnet2+ewald | net-charged | F | 2560 | 1286 | 4.43e-07 | 3.33e-07 | n/a | 264→264 | PASS |
| aimnet2+ewald | net-charged | F+S | 2560 | 1286 | 2.21e-07 | 3.88e-07 | 2.51e-05 | 264→264 | PASS |

Both matrices ran with TF32 disabled (measured matmul relative error 1.33e-06,
i.e. true fp32). Every partition is non-degenerate — each rank owns roughly half
the system — so the cross-rank reduction is genuinely exercised.

**On reproducing these numbers.** The per-row values are not stable run to run:
GPU reduction order varies, and these are error metrics sitting at the fp32
noise floor. 

### Additional multi-GPU gates

- Stress equivalence for Ewald and PME, slab Ewald and PME, MACE + cuEquivariance,
  and the wired AIMNet2→PME pipeline: 11 passed.
- AIMNet2 charge equivalence under DD, eager and compiled: slope `1.0000`,
  R² `1.0000`, max |Δq| `4.2e-07` across all 4 ranks.
- Compiled force-then-stress ordering: no divergence in any sequence,
  single-GPU or at world=4.

## Checklist

## Additional Notes

> [!TIP]

### DIFF
--- a/nvalchemi/distributed/_core/adapter.py
+++ b/nvalchemi/distributed/_core/adapter.py
@@ -549,38 +549,49 @@ _OUTPUT_TRANSFORM_REGISTRY: dict[str, type] = {
 }
 
 
-def _arg_transform_to_dict(t: ArgTransform) -> dict[str, Any]:
-    for kind, cls in _ARG_TRANSFORM_REGISTRY.items():
+def _transform_to_dict(t: Any, registry: dict[str, type], label: str) -> dict[str, Any]:
+    """Encode a transform as its registry name plus its dataclass fields.
+
+    The fields are carried generically rather than per-class: a transform field
+    that changes behaviour — an adjoint rule, say — would otherwise revert to its
+    default on the far side of a process boundary while the name still matched.
+    """
+    import dataclasses  # noqa: PLC0415
+
+    for kind, cls in registry.items():
         if isinstance(t, cls):
-            return {"type": kind}
-    raise TypeError(f"unknown ArgTransform: {type(t).__name__}")
+            encoded = {"type": kind}
+            encoded.update(dataclasses.asdict(t))
+            return encoded
+    raise TypeError(f"unknown {label}: {type(t).__name__}")
+
+
+def _transform_from_dict(
+    d: dict[str, Any], registry: dict[str, type], label: str
+) -> Any:
+    """Inverse of :func:`_transform_to_dict`."""
+    cls = registry.get(d.get("type"))
+    if cls is None:
+        raise ValueError(
+            f"unknown {label} type {d.get('type')!r}; expected one of {list(registry)}."
+        )
+    return cls(**{k: v for k, v in d.items() if k != "type"})
+
+
+def _arg_transform_to_dict(t: ArgTransform) -> dict[str, Any]:
+    return _transform_to_dict(t, _ARG_TRANSFORM_REGISTRY, "ArgTransform")
 
 
 def _arg_transform_from_dict(d: dict[str, Any]) -> ArgTransform:
-    cls = _ARG_TRANSFORM_REGISTRY.get(d.get("type"))
-    if cls is None:
-        raise ValueError(
-            f"unknown ArgTransform type {d.get('type')!r}; expected one of "
-            f"{list(_ARG_TRANSFORM_REGISTRY)}."
-        )
-    return cls()
+    return _transform_from_dict(d, _ARG_TRANSFORM_REGISTRY, "ArgTransform")
 
 
 def _output_transform_to_dict(t: OutputTransform) -> dict[str, Any]:
-    for kind, cls in _OUTPUT_TRANSFORM_REGISTRY.items():
-        if isinstance(t, cls):
-            return {"type": kind}
-    raise TypeError(f"unknown OutputTransform: {type(t).__name__}")
+    return _transform_to_dict(t, _OUTPUT_TRANSFORM_REGISTRY, "OutputTransform")
 
 
 def _output_transform_from_dict(d: dict[str, Any]) -> OutputTransform:
-    cls = _OUTPUT_TRANSFORM_REGISTRY.get(d.get("type"))
-    if cls is None:
-        raise ValueError(
-            f"unknown OutputTransform type {d.get('type')!r}; expected one of "
-            f"{list(_OUTPUT_TRANSFORM_REGISTRY)}."
-        )
-    return cls()
+    return _transform_from_dict(d, _OUTPUT_TRANSFORM_REGISTRY, "OutputTransform")
 
 
 # ----------------------------------------------------------------------
--- a/nvalchemi/distributed/_core/adapter.py
+++ b/nvalchemi/distributed/_core/adapter.py
@@ -119,6 +119,7 @@ from __future__ import annotations
 
 import logging
 import sys
+import warnings
 from dataclasses import dataclass, field
 from typing import Any, Callable, Literal, Sequence
 
@@ -360,6 +361,19 @@ class OpAdapter:
         )
 
     @property
+    def all_reduce_replicated_outputs(self) -> tuple[int, ...]:
+        """Subset of :attr:`all_reduce_outputs` whose reduced value feeds a
+        replicated whole-system computation, so the adjoint passes the incoming
+        gradient through instead of reducing it again."""
+        return tuple(
+            sorted(
+                p
+                for p, t in self.output_transforms.items()
+                if isinstance(t, AllReduceSum) and t.replicated_consumer
+            )
+        )
+
+    @property
     def slice_outputs_owned(self) -> tuple[int, ...]:
         return tuple(
             sorted(
@@ -404,6 +418,7 @@ class OpAdapter:
             scatter_outputs=self.scatter_outputs,
             owned_slice_inputs=self.owned_slice_inputs,
             all_reduce_outputs=self.all_reduce_outputs,
+            all_reduce_replicated_outputs=self.all_reduce_replicated_outputs,
             gather_inputs_full=self.gather_inputs_full,
             slice_outputs_owned=self.slice_outputs_owned,
         )
@@ -1278,24 +1293,39 @@ def _replacement_qualname(fn: Callable | None) -> str | None:
 
 def _resolve_replacement(qualname: str | None) -> Callable | None:
     """Inverse of :func:`_replacement_qualname`. Best-effort: returns
-    ``None`` if the qualname doesn't resolve."""
-    if not qualname or ":" not in qualname:
-        return None
-    import importlib  # noqa: PLC0415
+    ``None`` if the qualname doesn't resolve.
 
-    mod_path, qual = qualname.split(":", 1)
-    try:
-        mod = importlib.import_module(mod_path)
-    except Exception:
+    An unresolved replacement installs as a no-op, so the model silently runs
+    uncorrected; that case warns and the caller is expected to rebind from a
+    live spec.
+    """
+    if not qualname:
         return None
-    obj: Any = mod
-    for part in qual.split("."):
-        if part.startswith("<"):
-            return None
-        obj = getattr(obj, part, None)
-        if obj is None:
-            return None
-    return obj
+    resolved = None
+    if ":" in qualname:
+        import importlib  # noqa: PLC0415
+
+        mod_path, qual = qualname.split(":", 1)
+        try:
+            obj: Any = importlib.import_module(mod_path)
+        except Exception:
+            obj = None
+        for part in qual.split(".") if obj is not None else ():
+            if part.startswith("<"):
+                obj = None
+                break
+            obj = getattr(obj, part, None)
+            if obj is None:
+                break
+        resolved = obj
+    if resolved is None:
+        warnings.warn(
+            f"Adapter replacement {qualname!r} did not resolve; the adapter will "
+            "install as a no-op. Rebind it from a live spec before use.",
+            UserWarning,
+            stacklevel=3,
+        )
+    return resolved
 
 
 # Registry mapping serialized "kind" → adapter class. Subclasses that
--- a/nvalchemi/distributed/_core/halo_types.py
+++ b/nvalchemi/distributed/_core/halo_types.py
@@ -62,23 +62,47 @@ class ParticleHaloConfig:
     partitioner: Any  # SpatialPartitioner at runtime
     mesh: Any  # DeviceMesh at runtime
 
-    # Computed in __post_init__
+    # Computed on demand; see the topology properties below.
     rank: int = field(init=False)
-    neighbor_ranks: list[int] = field(init=False)
-    _pbc_images: dict[tuple[int, int], list[torch.Tensor]] = field(
-        init=False, repr=False
-    )
+    _topology: dict[str, Any] = field(init=False, repr=False)
 
     def __post_init__(self) -> None:
         try:
             self.rank = self.mesh.get_local_rank()
         except Exception:
             self.rank = 0
+        self._topology = {}
 
-        self.neighbor_ranks = [
-            r for r in self.partitioner.get_neighbor_ranks(self.rank) if r != self.rank
-        ]
-        self._pbc_images = _compute_pbc_image_vectors(self.partitioner)
+    def _current_topology(self) -> dict[str, Any]:
+        """Peer list and lattice images for the partitioner's current geometry.
+
+        Derived rather than stored: a barostat contraction narrows every domain,
+        which can widen the ghost shell onto ranks further away, and a copy taken
+        at construction would keep exchanging with the old peers. Recomputed only
+        when the partitioner reports a new topology.
+        """
+        version = self.partitioner.topology_version
+        if self._topology.get("version") != version:
+            self._topology = {
+                "version": version,
+                "neighbor_ranks": [
+                    r
+                    for r in self.partitioner.get_neighbor_ranks(self.rank)
+                    if r != self.rank
+                ],
+                "pbc_images": _compute_pbc_image_vectors(self.partitioner),
+            }
+        return self._topology
+
+    @property
+    def neighbor_ranks(self) -> list[int]:
+        """Ranks this one exchanges ghosts with, for the current geometry."""
+        return self._current_topology()["neighbor_ranks"]
+
+    @property
+    def _pbc_images(self) -> dict[tuple[int, int], list[torch.Tensor]]:
+        """Lattice images offered per rank pair, for the current geometry."""
+        return self._current_topology()["pbc_images"]
 
     @property
     def pbc_shifts(
--- a/nvalchemi/distributed/_core/halo_types.py
+++ b/nvalchemi/distributed/_core/halo_types.py
@@ -119,7 +119,10 @@ def _compute_pbc_image_vectors(
     cell_matrix = partitioner.cell_matrix
     pbc = partitioner.pbc
     grid = partitioner.rank_grid
-    span = getattr(partitioner, "neighbor_span", lambda: (1, 1, 1))()
+    # No fallback: a partitioner without a span cannot describe how far its
+    # ghost shell reaches, and silently assuming one domain is the very
+    # under-send this function exists to prevent.
+    span = partitioner.neighbor_span()
 
     total_ranks = grid[0] * grid[1] * grid[2]
     for sender_rank in range(total_ranks):
--- a/nvalchemi/distributed/_core/halo_types.py
+++ b/nvalchemi/distributed/_core/halo_types.py
@@ -25,6 +25,7 @@ directly.
 
 from __future__ import annotations
 
+import itertools
 from dataclasses import dataclass, field
 from typing import Any
 
@@ -106,13 +107,19 @@ def _compute_pbc_image_vectors(
 
     Returns ``{(sender, receiver): [image_1, image_2, ...]}`` where each
     ``(3,)`` image contains integer-valued fractional lattice coefficients.
-    For a diagonal neighbor crossing D periodic boundaries there are
-    ``2^D - 1`` independent images (all non-empty subsets of crossed dims).
+
+    A ghost shell wider than one domain reaches the same neighbor from both
+    sides of a periodic axis, and the two sides are different lattice images of
+    that neighbor holding different atoms. The offsets the shell actually spans
+    are therefore enumerated per axis rather than only the outermost pair; an
+    offset that needs no wrap is the direct (unshifted) copy the caller handles
+    separately, so only wrapping offsets appear here.
     """
     images: dict[tuple[int, int], list[torch.Tensor]] = {}
     cell_matrix = partitioner.cell_matrix
     pbc = partitioner.pbc
     grid = partitioner.rank_grid
+    span = getattr(partitioner, "neighbor_span", lambda: (1, 1, 1))()
 
     total_ranks = grid[0] * grid[1] * grid[2]
     for sender_rank in range(total_ranks):
@@ -120,36 +127,42 @@ def _compute_pbc_image_vectors(
         for receiver_rank in partitioner.get_neighbor_ranks(sender_rank):
             receiver_coords = partitioner.rank_to_grid_coords(receiver_rank)
 
-            per_dim_images: list[torch.Tensor] = []
+            # Per axis: which lattice translations reach this receiver at all.
+            axis_choices: list[tuple[int, list[int]]] = []
             for dim in range(3):
                 if not pbc[dim] or grid[dim] <= 1:
                     continue
-                dim_image = torch.zeros(
-                    3, device=cell_matrix.device, dtype=cell_matrix.dtype
-                )
-                if sender_coords[dim] == grid[dim] - 1 and receiver_coords[dim] == 0:
-                    dim_image[dim] = -1
-                elif sender_coords[dim] == 0 and receiver_coords[dim] == grid[dim] - 1:
-                    dim_image[dim] = 1
-                else:
-                    continue
-                per_dim_images.append(dim_image)
+                coefficients: set[int] = set()
+                for offset in range(-int(span[dim]), int(span[dim]) + 1):
+                    stepped = sender_coords[dim] + offset
+                    if stepped % grid[dim] != receiver_coords[dim]:
+                        continue
+                    # Exact: ``stepped - receiver`` is a whole number of grids.
+                    crossings = (stepped - receiver_coords[dim]) // grid[dim]
+                    if crossings != 0:
+                        coefficients.add(-crossings)
+                if coefficients:
+                    axis_choices.append((dim, sorted(coefficients)))
 
-            if not per_dim_images:
+            if not axis_choices:
                 continue
 
-            n = len(per_dim_images)
+            # A neighbor across several axes needs every combination of their
+            # translations; the all-zero one is the direct copy.
             combo_images: list[torch.Tensor] = []
-            for mask in range(1, 1 << n):
+            dims = [dim for dim, _ in axis_choices]
+            for choice in itertools.product(*([0, *c] for _, c in axis_choices)):
+                if not any(choice):
+                    continue
                 combo = torch.zeros(
                     3, device=cell_matrix.device, dtype=cell_matrix.dtype
                 )
-                for bit in range(n):
-                    if mask & (1 << bit):
-                        combo = combo + per_dim_images[bit]
+                for dim, coefficient in zip(dims, choice):
+                    combo[dim] = coefficient
                 combo_images.append(combo)
 
-            images[(sender_rank, receiver_rank)] = combo_images
+            if combo_images:
+                images[(sender_rank, receiver_rank)] = combo_images
 
     return images
 
--- a/nvalchemi/distributed/_runtime.py
+++ b/nvalchemi/distributed/_runtime.py
@@ -16,9 +16,12 @@
 
 from __future__ import annotations
 
+import contextlib
 import logging
 import os
 import warnings
+from collections.abc import Iterator
+from typing import Any
 
 import torch
 from physicsnemo.distributed import (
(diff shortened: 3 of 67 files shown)


nvalchemi/distributed/_core/halo_types.py
```diff
@@ -106,50 +107,62 @@ def _compute_pbc_image_vectors(
...
+    that neighbor holding different atoms. The offsets the shell actually spans
+    are therefore enumerated per axis rather than only the outermost pair; an
+    offset that needs no wrap is the direct (unshifted) copy the caller handles
+    separately, so only wrapping offsets appear here.
     """
     images: dict[tuple[int, int], list[torch.Tensor]] = {}
     cell_matrix = partitioner.cell_matrix
     pbc = partitioner.pbc
     grid = partitioner.rank_grid
+    span = getattr(partitioner, "neighbor_span", lambda: (1, 1, 1))()
```
laserkelvin: I'd prefer splitting this up to more than one line:

```python
span_method = getattr(partitioner, "neighbor_span", None)
if not span_method and not callable(span_method):
    span = (1, 1, 1)
else:
    span = span_method()
```

nvalchemi/distributed/validate/worker.py
```diff
@@ -49,6 +49,47 @@
 __all__ = ["_worker_main", "_patch_physicsnemo_all_to_all_for_gloo"]
 
 
+def _rebind_replacements(spec: MLIPSpec, live: "MLIPSpec | None") -> None:
```
laserkelvin: Why is `live` not type hinted as `MLIPSpec | None`?

nvalchemi/distributed/_runtime.py
```diff
@@ -75,3 +79,72 @@ def collective_device(fallback: torch.device | str = "cpu") -> torch.device:
...
+    for each and the results separate by far more than fp32 rounding. Also sets
+    ``NVIDIA_TF32_OVERRIDE``, which is what reaches ``mp.spawn`` / ``torchrun``
+    workers — they inherit the environment, not the torch flags. Call before the
+    process builds a CUDA context.
+
+    Returns
+    -------
+    None
+    """
+    os.environ.setdefault("NVIDIA_TF32_OVERRIDE", "0")
```
laserkelvin, line 106: Can we design this to be a context manager? If it means a substantial redesign we can mark it as a TODO, but IMO we should reset these variables after the context

nvalchemi/distributed/_runtime.py
```diff
@@ -75,3 +79,72 @@ def collective_device(fallback: torch.device | str = "cpu") -> torch.device:
...
+    torch.backends.cuda.matmul.allow_tf32 = False
+    torch.backends.cudnn.allow_tf32 = False
+    torch.set_float32_matmul_precision("highest")
+    for holder in (torch.backends.cuda.matmul, torch.backends.cudnn):
+        if hasattr(holder, "fp32_precision"):
+            try:
+                holder.fp32_precision = "ieee"
+            except Exception:  # pragma: no cover - varies by torch version
+                logger.debug("could not set fp32_precision", exc_info=True)
```
laserkelvin: This would continue even if it fails - should we re-raise the exception?

nvalchemi/distributed/_runtime.py
```diff
@@ -75,3 +79,72 @@ def collective_device(fallback: torch.device | str = "cpu") -> torch.device:
...
+
+
+def _is_reduced_precision(device: "Any" = None) -> bool:
```
laserkelvin: Device should be `str | torch.device | None`

nvalchemi/distributed/distributed_model.py
```diff
@@ -165,6 +165,12 @@ def _prepare_dd_compile(spec: "Any", compile_kwargs: "dict | None") -> dict:
         import torch._functorch.config as _fcfg  # noqa: PLC0415
 
         _fcfg.activation_memory_budget = float(_actb)
+    # A wired consumer backwards through this graph twice; AOT's donated-buffer
+    # optimisation frees buffers the second pass needs.
+    import torch._functorch.config as _fcfg2  # noqa: PLC0415
+
+    _fcfg2.donated_buffer = False
```
laserkelvin, line 172: Similar to the `pin_to_fp32` thing, maybe it would be good to have a collective context manager to handles all of these distributed settings, and cleans up after itself

nvalchemi/distributed/distributed_pipeline.py
```diff
@@ -367,48 +413,161 @@ def _run_wired_autograd_group(
...
-                        )
+                        cons_stress = cons_out.get("stress")
+                        # Either model may read a pinned field, so sum both backwards.
+                        _extra = [n for n in plan["grad_fields"] if n != "positions"]
+                        _fw = cons_out.get("_extra_grads")
+                        if _fw:
+                            # Reuse the framework's dE/dfield: a second backward
+                            # through a compiled graph returns a wrong chain term.
+                            de_dfield = _fw[0]
+                            cons_extra_grads = (
```
laserkelvin: Would be good if we could split this up into multiple lines - pretty high cognitive complexity

nvalchemi/distributed/partitioner.py
```diff
@@ -24,6 +24,10 @@
 
 from nvalchemi.distributed.config import DomainConfig
 
+# Relative slack when deciding how many domains a ghost shell spans, absorbing
+# float32 round-off in the derived domain width.
+_SPAN_ROUNDING_TOL = 1e-6
```
laserkelvin, line 29: Can we consolidate these hard coded values in the distributed module somewhere?

nvalchemi/distributed/particle_halo.py
```diff
@@ -45,6 +46,34 @@
...
+    bidx = padded.batch_idx
+    bidx = (bidx.to_local() if hasattr(bidx, "to_local") else bidx).long()
+    cell = sharded.cell
+    cell = cell.to_local() if hasattr(cell, "to_local") else cell
+
+    scaled_pos, scaled_cell = apply_strain(
+        atoms["positions"], cell, bidx, strain_pos, strain_cell
+    )
+    atoms["positions"] = scaled_pos
+    object.__setattr__(padded, "cell", scaled_cell)
```
laserkelvin: I'm unclear about the intention/behavior of this line; `object.__setattr__`, instead of `setattr(padded, "cell", scaled_cell)`?

nvalchemi/distributed/strategy.py
```diff
@@ -620,6 +620,22 @@ def _graph_partition_run_forward(
...
+        from nvalchemi.models._utils import prepare_strain  # noqa: PLC0415
+
+        cell_local = owned.cell
+        cell_local = (
+            cell_local.to_local() if hasattr(cell_local, "to_local") else cell_local
+        )
+        atoms["positions"], strained_cell, strain = prepare_strain(
+            pos, cell_local, owned.batch_idx.long()
+        )
+        object.__setattr__(owned, "cell", strained_cell)
```
laserkelvin: Same comment at another location; not sure if `setattr(owned, ...)` can be used instead

nvalchemi/distributed/partitioner.py
```diff
@@ -115,14 +120,27 @@ def update_cell(self, cell_matrix: torch.Tensor) -> None:
...
+        -------
+        None
         """
         cm = cell_matrix.squeeze(0) if cell_matrix.ndim == 3 else cell_matrix
+        previous_span = getattr(self, "_span", None)
         self.cell_matrix = cm.detach()
         self._inv_cell = torch.linalg.inv(self.cell_matrix)
+        self._span = self.neighbor_span()
+        if self._span != previous_span:
+            self._neighbor_ranks = self._compute_all_neighbor_ranks()
```
laserkelvin, line 148: `blocking:` `update_cell()` refreshes `_neighbor_ranks`, but the live `ParticleHaloConfig` keeps its construction-time `neighbor_ranks` and `_pbc_images`, so a contraction that widens `_span` still exchanges with the old peers.

Assuming barostat updates must make the halo match the current cell, refresh the halo topology atomically here (or derive it from the partitioner at exchange time) and add a contraction-after-halo-construction regression.

nvalchemi/distributed/_core/adapter.py
```diff
@@ -359,6 +360,19 @@ def all_reduce_outputs(self) -> tuple[int, ...]:
...
 
+    @property
+    def all_reduce_replicated_outputs(self) -> tuple[int, ...]:
+        """Subset of :attr:`all_reduce_outputs` whose reduced value feeds a
+        replicated whole-system computation, so the adjoint passes the incoming
+        gradient through instead of reducing it again."""
+        return tuple(
+            sorted(
+                p
+                for p, t in self.output_transforms.items()
```
laserkelvin, line 371: `blocking:` `replicated_consumer` changes the all-reduce adjoint, but `OpAdapter.to_dict()` serializes this transform as only `{"type": "all_reduce_sum"}`; `from_dict()` restores the default `False`.

Assuming spawned validator and production workers must preserve the same backward rule, serialize this flag and cover a `True` round trip.

nvalchemi/distributed/distributed_pipeline.py
```diff
@@ -367,48 +413,161 @@ def _run_wired_autograd_group(
...
+                )
+                cons_stress = None
                 try:
-                    with DistributedModel(consumer, plan["consumer_cfg"]) as cdm:
+                    # ``extra_grad_inputs`` keeps the graph alive for this
+                    # group's backward.
+                    cons_compiled = self._compile and self._compile_capable(
+                        plan["consumer"]
+                    )
+                    with DistributedModel(
```
laserkelvin: `blocking:` the wired path constructs and closes both `DistributedModel`s inside every forward, so the consumer loses capacity growth and its per-instance compiled-region cache each MD step; the producer also ignores the pipeline compile setting.

Assuming compiled wired runs should reach the documented persistent steady state, build both models when planning the group, retain them, and close them with the pipeline.

nvalchemi/distributed/sharded_batch.py
```diff
@@ -540,6 +641,12 @@ def local_batch_with_edges(
                 ctor_kwargs[name] = tensor
             else:
                 extras[name] = tensor
+        system_extras: dict[str, torch.Tensor] = {}
+        for name, tensor in self.system_fields.items():
```
laserkelvin, line 645: `blocking:` this restores generic `system_fields` for local and gathered batches, but `HaloStrategy.apply_migration()` rebuilds a separate `Batch` and copies only `cell`, `pbc`, `energy`, and `stress`. After migration, total charge, spin, and custom system properties disappear from the returned batch.

Assuming migration changes ownership without changing the system schema, route migration through this same system-field packing path and add a migration regression.

laserkelvin (commented):
I found four blocking correctness and lifecycle issues in the new halo, serialization, wired compilation, and migration paths; details are inline.

nvalchemi/models/_ops/electrostatics/ewald_recip_torch.py
```diff
@@ -137,6 +137,15 @@ def ewald_energy_from_structure_factors(
     :func:`ewald_partial_structure_factors` + a cross-rank all-reduce the weighted
     VJP is exact for any cotangent (incl. the owned-mask DD consolidation
     produces). Returns per-atom energy ``(N,)`` float64.
+
+    ``charges`` is cast to float64 ONCE and that tensor is used for both the
+    k-sum and the corrections call. Passing the caller's original (float32)
+    tensor to the corrections while the k-sum uses the cast copy reaches the
+    opaque kernel from one leaf by two routes of differing dtype, which
+    ``torch.compile`` mis-accumulates: the traced forward and ``dE/dpositions``
+    stay exact while ``dE/dcharges`` comes out ~28% wrong (and near-zero in a
```
laserkelvin, line 147: Oddly specific 28%

nvalchemi/models/_utils.py
```diff
@@ -179,29 +180,64 @@ def prepare_strain(
         The returned tensor is an unconstrained autograd leaf; only its
         symmetric part is applied as strain.
     """
-    n_systems = cell.shape[0]
     displacement = torch.zeros(
-        n_systems,
+        cell.shape[0],
```
laserkelvin: Why did this go from `n_systems` to `cell.shape[0]`?

nvalchemi/models/ewald.py
```diff
@@ -203,6 +208,16 @@ def distribution_spec(self, strategy: Any = None) -> Any:
         if strategy == StrategyKind.GRAPH_PARTITION:
             return self._distribution_spec_gp()
 
+        if self.hybrid_forces:
+            raise NotImplementedError(
+                "Ewald under domain decomposition requires hybrid_forces=False. "
```
laserkelvin, line 217: This assumes that the user is running this under domain decomposition

nvalchemi/models/pme.py
```diff
@@ -235,6 +235,16 @@ def distribution_spec(self, strategy: Any = None) -> Any:
         if strategy == StrategyKind.GRAPH_PARTITION:
             return self._distribution_spec_gp()
 
+        if self.hybrid_forces:
+            raise NotImplementedError(
+                "PME under domain decomposition requires hybrid_forces=False. "
```
laserkelvin, line 244: Same comment as in Ewald

test/distributed/_core/test_compile_per_system.py
```diff
@@ -33,7 +33,7 @@
 
 def _worker_per_system(rank, world_size):
     os.environ["MASTER_ADDR"] = "127.0.0.1"
-    os.environ["MASTER_PORT"] = "29681"
+    os.environ["MASTER_PORT"] = "29709"
```
laserkelvin, line 36: Odd that this needed to change

test/distributed/model/test_distributed_models.py
```diff
@@ -717,9 +720,13 @@ def _mace_wrapper(dtype: torch.dtype, device: str = "cpu"):
...
-    """Deterministic free-ish port per-parametrization so concurrent
-    pytest runs don't collide. 30000 + hash(key)%5000."""
-    return str(30000 + (hash(key) & 0xFFFF) % 5000)
+    """Deterministic per-parametrization rendezvous port.
+
+    Kept below the kernel's ephemeral range: a port drawn from there can be
+    claimed for an outbound connection between the choice and the rendezvous
+    bind, which surfaces as a mid-test EADDRINUSE.
+    """
+    return str(20000 + (hash(key) & 0xFFFF) % 9000)
```
laserkelvin, line 729: This sets a rendezvous port but isn't necessarily free:

```python
from socket import socket

with socket() as sock:
    sock.bind(('', 0))
    _, free_port: int = sock.getsockname()[1]
```

nvalchemi/distributed/_core/adapter.py
dallasfoster: Confirmed, and thank you — this is the same failure this PR opens with, one level further down.

```python
>>> t = AllReduceSum(replicated_consumer=True)
>>> _output_transform_to_dict(t)
{'type': 'all_reduce_sum'}
>>> _output_transform_from_dict(_)
AllReduceSum(replicated_consumer=False)
```

The encoder emitted only `{"type": kind}` and the decoder called `cls()`, so the flag reverted to its default and the backward rule changed across any process boundary while the transform's name still matched. I audited the spec's own fields and never looked inside the transforms nested in `custom_ops`.

Fixed in `dc9b7e10`, generically rather than per-field: transforms are frozen dataclasses, so the encoder now carries their fields and the decoder replays them — a field added later is covered without anyone remembering. The guard test walks both registries and round-trips a non-default instance of each; I verified it fails with the old encoder.

It also caught a pre-existing test asserting the buggy shape (`{"type": "all_reduce_sum"}` with no fields), which I updated, since that expectation was encoding the defect.

nvalchemi/distributed/partitioner.py
dallasfoster: Confirmed. Reproduced by contracting a 48 Å box to 16 Å at world=4:

| | span | peers of rank 1 | images cached for (1, 3) |
|---|---|---|---|
| at construction | (1, 0, 0) | — | 0 |
| after `update_cell` | (2, 0, 0) | [0, 2, 3] | **0** |

The partitioner correctly learns it must now reach rank 3 and the halo config never hears about it, so the exchange keeps the old peer set and the old lattice images. Both were built once in `__post_init__`.

This had made the span work in this PR only half-complete: it computed the right span, and a barostat run never picked it up.

Fixed in `972d5ff6` by deriving rather than invalidating — the partitioner carries a topology version it bumps when the neighbour set changes, and the halo config recomputes when it sees a new one. A cache that has to be invalidated from a second object is exactly how this happened, so the version lives with the data it describes.

Regression added (`TestContractionRefreshesHaloTopology`), verified to fail without the fix.

nvalchemi/distributed/sharded_batch.py
dallasfoster: Confirmed, and it is a hole in this PR's own fix.

`apply_migration` rebuilds via `_build_batch_from_fields` (per-atom only) and then restored per-system state by naming `cell`, `pbc`, `energy` and `stress` individually. Everything else attached to the system group — total `charge`, `spin`, anything added through `add_system_property` — was gone after the first migration step. An AIMNet2 run under a barostat would silently revert to a neutral system partway through a trajectory.

One of the reported bugs this PR fixes is exactly "system-level inputs dropped"; we routed the local and gathered paths through `system_fields` and missed migration as a third path.

Fixed in `ba4e0e8e` by carrying the whole system group: migration changes atom ownership, not the system schema, so every per-system tensor crosses verbatim. Verified `charge` and a custom property now survive where both were previously dropped, with a regression in `test_system_fields.py`.

nvalchemi/distributed/distributed_pipeline.py
dallasfoster: Confirmed. Both models are constructed inside `_run_wired_autograd_group` (lines 374 and 438; the method spans 312–567), so every MD step built and tore down both `DistributedModel`s, discarding the shape caps they had grown and the per-instance compiled-region cache. A compiled wired run therefore recompiled every step — the opposite of the recompile-free steady state the single-model gates assert.

Fixed in `2226a115` as you describe: both are built when the group is planned, as the per-step path already did, and closed with the pipeline.

Two caveats I would rather state than have you find:

**This is structurally verified but not behaviourally validated here.** The pipeline tests run 1 passed / 8 skipped locally — the wired path is almost entirely multi-GPU gated. The restructuring dedented two `with` bodies inside nested `try`/`finally`, so I checked via AST that both `try` blocks kept their `finally` and no orphaned `with` remains, but that is not the same as running it. A multi-GPU job is queued. You are also right that the wired path deserves a recompile gate of its own — the existing one covers single models, which is why this went unnoticed.

**I deliberately left the producer-compile half alone.** Enabling it routes the wired producer onto the framework-autograd path, and there is an existing constraint that the producer runs derivative-free — otherwise the graph the surrogate backward walks gets freed. That is a correctness question I would rather answer with evidence than change blind. Happy to take it as a follow-up.

nvalchemi/distributed/_core/halo_types.py
dallasfoster: I ended up deleting instead of splitting. It's unreachable.

nvalchemi/distributed/validate/worker.py
dallasfoster: Fixed.

nvalchemi/distributed/_runtime.py
dallasfoster: Agreed, added as `pinned_fp32()`.

nvalchemi/distributed/distributed_model.py
dallasfoster: Leave as a follow-up.

nvalchemi/distributed/distributed_pipeline.py
dallasfoster: Agreed — split into statements. The `dict(zip(...))` around an `autograd.grad` inside
a conditional expression is now a plain `if` with the grad call bound to a local.

nvalchemi/distributed/partitioner.py
dallasfoster: Probably at some point, it's a little unclear where that would be at the moment.

nvalchemi/models/_ops/electrostatics/ewald_recip_torch.py
dallasfoster: Fixed.

nvalchemi/models/_utils.py
dallasfoster: No good reason

nvalchemi/models/ewald.py
dallasfoster: This lies within the distributed spec method, so it's a fine assumption.

nvalchemi/models/pme.py
dallasfoster: Same response, it's actually fine.

test/distributed/model/test_distributed_models.py
dallasfoster: Really this is more complicated than it appears, at least on certain H100 clusters.

test/distributed/_core/test_compile_per_system.py
dallasfoster: Ran into bugs on a certain cluster.

nvalchemi/distributed/_runtime.py
dallasfoster: Fixed.

nvalchemi/distributed/_runtime.py
dallasfoster: Not so much of a redesign, but yes we should mark as TODO.

dallasfoster:
/ok to test [52074f3](https://github.com/NVIDIA/nvalchemi-toolkit/pull/160/commits/52074f3cb6f059ca416c4700937ab706a9ec5f53)

dallasfoster:
/ok to test [b825f42](https://github.com/NVIDIA/nvalchemi-toolkit/pull/160/commits/b825f4252033a32f424dfd749b681f93a5cabcd5)

PR #168, opened 2026-08-13
feat(dynamics): support compiled fused steps with CUDA graphs

laserkelvin:
## Description

Make `FusedStage.step()` compatible with `torch.compile` and CUDA graphs, and coordinate Torch and Warp work on the same CUDA stream.

Previously, lazy allocations and data-dependent branches inside the fused step prevented full-graph capture, while Warp could launch on a separate stream from Torch and invalidate CUDA-graph capture. This change moves setup outside the compiled region, uses static-shape masked updates, and introduces a joint stream context for both runtimes.

## Type of Change

- [x] Performance improvement
- [x] Documentation update

## Changes Made

- Restructure the fused dynamics step to avoid lazy allocation and data-dependent control flow in the compiled region.
- Keep autograd inputs and masked updates capture-safe across repeated CUDA-graph replays.
- Run Torch and Warp kernels on one dedicated stream when dynamics is used as a context manager, while binding Warp to the caller's current stream for bare runs.
- Add regression coverage for compilation, masked updates, stream propagation, and context-manager behavior.
- Document the CUDA stream behavior in the dynamics API guidance.

## Testing

- [x] `uv run --extra cu13 pytest test/dynamics/test_single_loop.py -q` (87 passed on NVIDIA GB10)
- [x] Pre-commit hooks pass for all changed files
- [x] New regression tests added
- [ ] Full test suite not run

## Checklist

- [x] I have read and understand the contributing guidelines.
- [ ] `CHANGELOG.md` was not updated.
- [x] I have performed a self-review of the changes.
- [x] Documentation was updated where applicable.

### DIFF
--- a/nvalchemi/dynamics/base.py
+++ b/nvalchemi/dynamics/base.py
@@ -794,10 +794,11 @@ class _CommunicationMixin:
         it jointly on the torch side (``torch.cuda.StreamContext``) and the
         warp side (``wp.ScopedStream`` over the converted stream), so all
         subsequent GPU operations — including warp-backed custom ops —
-        execute on the one dedicated stream.  On non-CUDA devices this is
-        a no-op.  See the **CUDA stream semantics** notes on
-        :class:`BaseDynamics` for how this differs from calling ``run()``
-        without the context manager.
+        execute on the one dedicated stream. The dedicated stream first waits
+        for work already submitted to the caller's current torch stream. On
+        non-CUDA devices this is a no-op. See the **CUDA stream semantics**
+        notes on :class:`BaseDynamics` for how this differs from calling
+        ``run()`` without the context manager.
 
         Returns
         -------
@@ -805,7 +806,9 @@ class _CommunicationMixin:
             This instance.
         """
         if self.device_type == "cuda" and torch.cuda.is_available():
+            caller_stream = torch.cuda.current_stream(self.device)
             self._stream = torch.cuda.Stream(device=self.device)
+            self._stream.wait_stream(caller_stream)
             self._wp_stream = wp.stream_from_torch(self._stream)
             self._stream_ctx = self._enter_joint_stream_context(
                 self._stream, self._wp_stream
--- a/nvalchemi/dynamics/base.py
+++ b/nvalchemi/dynamics/base.py
@@ -124,6 +124,7 @@ def requires_grad_ctx(
     """
     unique: list[torch.Tensor] = []
     seen: set[int] = set()
+    # perform some intial validation of the data
     for tensor in tensors:
         if not isinstance(tensor, torch.Tensor):
             raise TypeError(f"Expected a tensor, got {type(tensor).__name__}")
@@ -138,6 +139,8 @@ def requires_grad_ctx(
             unique.append(tensor)
 
     original = [(tensor, tensor.requires_grad) for tensor in unique]
+    # set gradient tracking, release control flow back, and when it
+    # wraps up, rest to the original tracking state
     try:
         for tensor, state in original:
             if state != requires_grad:
--- a/nvalchemi/dynamics/base.py
+++ b/nvalchemi/dynamics/base.py
@@ -52,6 +52,7 @@ import sys
 import warnings
 from collections import OrderedDict
 from collections.abc import Callable, Sequence
+from contextlib import ExitStack, nullcontext
 from enum import Enum
 from typing import (
     TYPE_CHECKING,
@@ -62,6 +63,7 @@ from typing import (
 )
 
 import torch
+import warp as wp
 from jaxtyping import Bool
 from loguru import logger
 from pydantic import BaseModel, ConfigDict, Field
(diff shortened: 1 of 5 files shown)


nvalchemi/dynamics/base.py
```diff
@@ -3125,11 +3292,8 @@ def _step_impl(self, batch: Batch) -> tuple[Batch, torch.Tensor | None]:
         newly_graduated = (pre_migration_status < self.exit_status) & (
             post_status >= self.exit_status
         )
-        exit_converged: torch.Tensor | None = (
-            torch.where(newly_graduated)[0] if newly_graduated.any() else None
-        )
 
-        return batch, exit_converged
+        return batch, newly_graduated
```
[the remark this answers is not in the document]
laserkelvin: Agreed. We will address convergence compilation in a follow-up and keep it out of scope here.

nvalchemi/dynamics/base.py
```diff
@@ -1353,6 +1460,16 @@ class BaseDynamics(HookRegistryMixin, _CommunicationMixin):
 
     Notes
     -----
+    **CUDA streams.** Used as a context manager, dynamics creates a dedicated
+    stream shared by torch and warp. This supports overlapping GPU work,
+    multi-stage pipelines, and CUDA-graph capture. Entry waits for prior work,
```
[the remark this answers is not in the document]
laserkelvin: Agreed. Fixed in 7b02ac6 by making the dedicated stream wait on the current Torch stream before entry.

laserkelvin:
/ok to test [7b02ac6](https://github.com/NVIDIA/nvalchemi-toolkit/pull/168/commits/7b02ac6956b7a3c30db4fe6ca2f103f951c153b5)

laserkelvin:
/ok to test [c4c479f](https://github.com/NVIDIA/nvalchemi-toolkit/pull/168/commits/c4c479fb45961bf439689fb6dcb2992f43ec464b)

nvalchemi/dynamics/base.py
dallasfoster, line 1: So if I run the following code:

```python
# Use model as a thin DemoModel
fused = FusedStage(..., compile_step = True, compile_kwargs = 'reduce_overhead')
batch = Batch.from_data_list(...)

with fused:
    for _ in range(8):
        fused.step(batch)
torch.cuda.synchronize()

print("cudagraph_skips:", counters["inductor"]["cudagraph_skips"])
```
then on `main` I get zero skips but on this branch I get four because of `val.copy_` and `target.copy_` operations.

dallasfoster: I'm observing the following behavior:
```text
BaseDynamics.run -> positions.requires_grad = False
FusedStage.run   -> positions.requires_grad = True
   positions.numpy()   RuntimeError: Can't call numpy() on Tensor that requires grad.
```

this also causes cudagraph refusal's "outputs from a previous step still require backward".

test/dynamics/test_single_loop.py
dallasfoster, line 1: I suggest you add a test that executes  a compiled step. Every compile test either asserts `_compiled_step is not None` (lines 543, 693, 773) or mocks `torch.compile` (line 713), and the new stream tests mock `wp` entirely.

nvalchemi/dynamics/base.py
```diff
@@ -2241,29 +2387,38 @@ def _masked_post_update(
...
-                continue
-            if val.shape[0] == batch.num_nodes:
-                saved[field] = val[~node_mask].clone()
-            elif val.shape[0] == batch.num_graphs:
-                saved[field] = val[sys_mask].clone()
-
-        self.post_update(batch)
+            if val is not None:
+                saved[field] = val.clone()
+        return saved
```
dallasfoster, line 2467: Consider using `torch.where(..., out=val)`

dallasfoster (changes requested):

nvalchemi/dynamics/base.py
laserkelvin: For the first, I've added a unit test and in the second, added a context manager for managing `requires_grad`. Potentially we can re-use the same context manager in other parts of the codebase as well

test/dynamics/test_single_loop.py
laserkelvin: The unit test I added as part of your other comment checks by running a few steps, and asserts zero skips

nvalchemi/dynamics/base.py
laserkelvin: Done in [dfb01fe](https://github.com/NVIDIA/nvalchemi-toolkit/pull/168/commits/dfb01fe36bd8619ac0524630e3c4e180818348ff)

laserkelvin:
/ok to test [d4f7327](https://github.com/NVIDIA/nvalchemi-toolkit/pull/168/commits/d4f7327cc066d2e78b15d96b5519190bb0c219b3)

nvalchemi/dynamics/base.py
dallasfoster: It seems we still have a val.clone() that does full tensor clones too.

dallasfoster (approved):

PR #181, opened 2026-08-26
fix: make dynamics hook lifecycle status-aware

ys-teh:

## Description

Make dynamics hook execution respect each dispatch's active graphs, align the hook lifecycle across standalone, fused, and domain-parallel dynamics, and initialize forces before the first integration step.

There will be a separate PR to deprecate `FusedStage.fused_hooks`, revise the order in which the `FusedStage.hooks` and substage hooks are applied, and add clear documentation.

Items deferred for future work:
- The way `FreezeAtomsHook` handles frozen atoms may still not be sufficient for every integrator. For example, thermostats and barostats may need an explicit interface to exclude frozen degrees of freedom from their internal statistics and state update.
- Force priming on DistributedPipeline and potential implication on DomainParallel.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Related Issues

N/A

## Changes Made

- Add `active_graph_mask` to dynamics hook contexts and propagate it through `BaseDynamics`, `FusedStage`, and `DomainParallel` hook dispatch.
- Restrict mutating and validation hooks to active graphs, and fix `FreezeAtomsHook` so constrained positions, velocities, and forces are handled at the correct split-update stages.
- Prime model forces before integration and expose `AFTER_PRE_UPDATE`/`BEFORE_POST_UPDATE` consistently for fused substages.

## Testing

Targeted tests run:

- `uv run --extra cu13 pytest test/dynamics/ test/hooks/ test/distributed/test_domain_parallel.py test/distributed/test_domain_parallel_convergence.py` (passed: 1246 passed, 8 skipped)

Full-suite attempt:

- `make pytest` (not completed: the first attempt could not collect without the declared `ase` extra; the rerun was stopped in favor of the scoped dynamics, hooks, and DomainParallel suites above)

## Checklist

## Additional Notes

> [!TIP]
> of the code, nor is it an indication that the PR will be accepted/rejected.


nvalchemi/dynamics/base.py
```diff
@@ -2208,6 +2276,13 @@ def run(self, batch: Batch, n_steps: int | None = None) -> Batch:
...
         self._open_hooks()
         try:
+            status = getattr(batch, "status", None)
+            if status is None:
+                active_graph_mask = None
+            else:
+                status = status.squeeze(-1) if status.dim() == 2 else status
+                active_graph_mask = status[: batch.num_graphs] < self.exit_status
+            self._prime_forces(batch, active_graph_mask)
             for _ in range(resolved):
```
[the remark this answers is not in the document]
ys-teh: Added validation for n_steps to be non-negative. Ideally n_steps should be > 0. But when n_steps=0, it will run prime forces.

nvalchemi/dynamics/base.py
```diff
@@ -2154,17 +2190,49 @@ def step(self, batch: Batch) -> tuple[Batch, torch.Tensor | None]:
...
+                DynamicsStage.BEFORE_COMPUTE,
+                batch,
+                active_graph_mask,
+            )
+            self.compute(batch)
+            self._call_hooks(
+                DynamicsStage.AFTER_COMPUTE,
+                batch,
+                active_graph_mask,
+            )
```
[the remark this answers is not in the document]
ys-teh: This matches the existing FusedStage priming behavior, so we will keep it consistent in this PR and address step count separately.

nvalchemi/distributed/domain_parallel.py
```diff
@@ -610,8 +671,25 @@ def _gather_all(self, local_batch: Batch) -> Batch:
     # Hook overrides
     # ------------------------------------------------------------------
 
-    def _build_context(self, batch: Batch) -> HookContext:
-        ctx = super()._build_context(batch)
+    def _active_graph_mask(self, batch: Batch) -> torch.Tensor | None:
```
laserkelvin: Can you add consistent `jaxtyping` for the active graph mask?

nvalchemi/hooks/_context.py
```diff
@@ -67,10 +67,16 @@ class DynamicsContext(HookContext):
...
         ``None`` when convergence has not fired for this dispatch.
+    active_graph_mask : torch.Tensor | None
+        Boolean mask of shape ``(batch.num_graphs,)`` selecting graphs active in
+        the current dynamics dispatch. ``None`` when the dispatch has no status-based
+        filtering (typically standalone ``BaseDynamics``). In a ``FusedStage``
+        substage, it selects graphs whose status matches that substage.
     """
 
     step_count: int = 0
     converged_mask: torch.Tensor | None = None
```
laserkelvin: While you're at it, could you add `jaxtyping` shape annotations for `active_graph_mask`, and `converged_mask`?

nvalchemi/dynamics/base.py
```diff
@@ -3555,16 +3745,19 @@ def run(
...
             # this initial forward pass the first step would integrate with
             # zero (uninitialised) forces.
-            self._call_hooks(DynamicsStage.BEFORE_COMPUTE, batch)
-            self.compute(batch)
-            self._call_hooks(DynamicsStage.AFTER_COMPUTE, batch)
+            status = batch.status
+            if status.dim() == 2:
+                status = status.squeeze(-1)
+            active_graph_mask = status[: batch.num_graphs] < self.exit_status
+            self._prime_forces(batch, active_graph_mask)
```
laserkelvin: `blocking:` This primes only the initial batch. After `refill_check()` replaces all graphs, the new `Batch` can lack `forces` and `energy`, and the next loop calls `step()` without allocating or priming them. A minimal repro with six two-atom samples, `max_batch_size=3`, `refill_frequency=1`, and immediate convergence calls `_prime_forces` once, then fails on the replacement with `KeyError: expected='forces'`. Please allocate required output buffers after refill, prime the replacement before its first `pre_update`, and add this case to `test_inflight.py`.

nvalchemi/dynamics/hooks/cell_align.py
```diff
@@ -113,5 +119,13 @@ def __call__(self, ctx: DynamicsContext, stage: Enum) -> None:
...
 
-        batch.positions = positions
-        batch.cell = cell
+        if active_graph_mask is None:
+            batch.positions = positions
+            batch.cell = cell
+            return
+
+        # Update results only for active graphs, leaving inactive graphs unchanged
+        active_atoms = active_graph_mask[batch.batch_idx].unsqueeze(-1)
```
laserkelvin: `blocking:` The final write uses only `active_graph_mask`, so it drops the periodic check above. With `pbc=[[True, True, True], [False, False, False]]` and both graphs active, this hook changes the second graph's cell and positions even though it is nonperiodic. Nonperiodic graphs should remain unchanged. Please add a mixed-PBC unit test.

nvalchemi/distributed/domain_parallel.py
```diff
@@ -354,14 +379,22 @@ def step(self, batch: Batch) -> tuple[Batch, torch.Tensor | None]:
...
 
     # ------------------------------------------------------------------
     # Force priming (initial compute before the first integrator step)
     # ------------------------------------------------------------------
 
-    def _prime_forces(self, batch: Batch) -> None:
+    def _prime_forces(
+        self,
+        batch: Batch,
+        active_graph_mask: torch.Tensor | None,
```
laserkelvin, line 396: `blocking:` Making `active_graph_mask` required leaves `_test_prime_forces` calling `dd._prime_forces(local_batch)` with the old signature. The two-GPU test will raise `TypeError` before it checks force priming. Please pass `dd._active_graph_mask(local_batch)` at that call site and run the multigpu test.

nvalchemi/dynamics/base.py
```diff
@@ -2206,8 +2285,16 @@ def run(self, batch: Batch, n_steps: int | None = None) -> Batch:
...
+        self._validate_n_steps(resolved)
         self._open_hooks()
         try:
+            status = getattr(batch, "status", None)
+            if status is None:
+                active_graph_mask = None
+            else:
+                status = status.squeeze(-1) if status.dim() == 2 else status
+                active_graph_mask = status[: batch.num_graphs] < self.exit_status
+            self._prime_forces(batch, active_graph_mask)
```
laserkelvin: `blocking:` `run()` now initializes forces, but public `step()` does not, so the two entry points produce different first steps. Minimal repro: clone one batch with zero initial forces, call `DemoDynamics.step()` on one copy and `DemoDynamics.run(..., n_steps=1)` on the other; the positions and velocities differ. Since `run()` is documented as repeatedly calling `step()`, the first step should match. Please move force readiness into a shared first-step path and add this equality test.

nvalchemi/distributed/domain_parallel.py
```diff
@@ -683,7 +778,10 @@ def run(self, batch: Batch, n_steps: int | None = None) -> Batch:
         self._open_hooks()
         try:
             if not self._forces_primed:
-                self._prime_forces(batch)
+                self._prime_forces(
```
laserkelvin: `issue:` This override never calls `_validate_n_steps(resolved)`. A small unit repro can force this branch with `_dist_model=object()` and `_forces_primed=True`; `run(batch, n_steps=-1)` then returns the unchanged batch with `step_count == 0` instead of raising. `DomainParallel` should follow the same input contract as `BaseDynamics` and `FusedStage`. Please validate `resolved` before opening hooks and add negative/bool tests.

laserkelvin (changes requested):
Requesting changes for five concrete lifecycle issues. Each inline comment includes a small reproducer and a focused fix.

laserkelvin:
Additionally want to probably update docs:

  - `docs/modules/dynamics/fused_stage.rst:73` still shows compute() followed by each stage’s combined masked_update(). PR 181 now performs masked pre_update, one shared compute, then masked post_update.
  - `docs/modules/dynamics/hooks.rst:319` says fused substages do not fire AFTER_PRE_UPDATE or BEFORE_POST_UPDATE. PR 181 adds both. The diagram below it also shows the old ordering.
  - `docs/modules/dynamics/implementing_dynamics.rst:223` teaches implementers that FusedStage calls one combined masked_update(). That is no longer how fused execution works.
  - `docs/modules/dynamics/fused_stage.rst:119` says `n_steps` is unused and fused termination is purely convergence-driven. PR 181 validates and uses n_steps as a maximum.
  - `docs/userguide/hooks.md:93` omits the new DynamicsContext.active_graph_mask field. It also still documents _build_context(batch) as the custom-context interface.
  - `docs/modules/dynamics/hooks.rst:242` describes FreezeAtomsHook as a two-stage hook. PR 181 makes it fire across five stages and changes when zero_forces=False exposes forces.
  - The rendered API documentation will also inherit a stale source docstring: `nvalchemi/dynamics/base.py:2890` still says fused execution computes first and then applies masked_update().

nvalchemi/distributed/domain_parallel.py
ys-teh: Added in https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/4afabe576ca1acf91d9cdd83c68fb72c542c7055

nvalchemi/hooks/_context.py
ys-teh: Added in https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/4afabe576ca1acf91d9cdd83c68fb72c542c7055

nvalchemi/dynamics/hooks/cell_align.py
ys-teh: Thanks for highlighting this. The old version (before active_graph_mask was added) also did not take into account the periodic mask, likely because it is rare to encounter cases with mixed boundary condition. I have updated it to account for the periodicity mask here regardless: https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/6891fd244385f19f96e9b9553605a5d849f00dc3

nvalchemi/distributed/domain_parallel.py
ys-teh: Added in https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/021aa6d7d30c18d31cd152c7fd0c6701dd615ecb

nvalchemi/distributed/domain_parallel.py
ys-teh: Updated in https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/021aa6d7d30c18d31cd152c7fd0c6701dd615ecb

nvalchemi/dynamics/base.py
ys-teh: Updated in https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/021aa6d7d30c18d31cd152c7fd0c6701dd615ecb with `self._forces_primed=False` so the primes will happen at the beginning of the next step.

nvalchemi/dynamics/base.py
ys-teh: I have now standardized the force priming behavior across BaseDynamics, FusedStage, and DomainParallel:
- self._forces_primed=False in .run before any .step.
- run self._prime_forces() and set self._forces_primed=True in the first step.
I also removed the extra force priming in .run in DomainParallel since it is going to be called anyway in .step.

nvalchemi/dynamics/base.py
ys-teh: Thanks for flagging this. It's now fixed. Relevant test: `test_refill_allocates_and_primes_replacement_outputs`.

I ended up making a bigger change because I realized graphs can enter new stages at different iterations. Repriming the whole batch every time would add an extra model evaluation to many iterations. So now graphs that need repriming will use one iteration to do that by skipping pre-update and post-update. [2de2e76](https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/2de2e76740e24cfca674b80a175e3140c4f53f8a)

In the same commit, I also fixed masked updates so inactive graphs won't change the sub-stage optimizer/integrator state, and added an optional `reprime_on_entry` for stages that need fresh model outputs before their first update.

One item to note is that `batch.velocities` may still be incorrect when a graph transitions between stages. Fixing this requires a larger refactor, so I’ve deferred it for now. As a temporary workaround, velocities can be adjusted with a hook.

ys-teh:
> Additionally want to probably update docs:
> 
> * `docs/modules/dynamics/fused_stage.rst:73` still shows compute() followed by each stage’s combined masked_update(). PR 181 now performs masked pre_update, one shared compute, then masked post_update.
> * `docs/modules/dynamics/hooks.rst:319` says fused substages do not fire AFTER_PRE_UPDATE or BEFORE_POST_UPDATE. PR 181 adds both. The diagram below it also shows the old ordering.
> * `docs/modules/dynamics/implementing_dynamics.rst:223` teaches implementers that FusedStage calls one combined masked_update(). That is no longer how fused execution works.
> * `docs/modules/dynamics/fused_stage.rst:119` says `n_steps` is unused and fused termination is purely convergence-driven. PR 181 validates and uses n_steps as a maximum.
> * `docs/userguide/hooks.md:93` omits the new DynamicsContext.active_graph_mask field. It also still documents _build_context(batch) as the custom-context interface.
> * `docs/modules/dynamics/hooks.rst:242` describes FreezeAtomsHook as a two-stage hook. PR 181 makes it fire across five stages and changes when zero_forces=False exposes forces.
> * The rendered API documentation will also inherit a stale source docstring: `nvalchemi/dynamics/base.py:2890` still says fused execution computes first and then applies masked_update().

Docs are deferred to #182 to prevent merge conflict.

test/distributed/model/test_multigpu.py
```diff
@@ -356,7 +356,7 @@ def _test_prime_forces(rank: int, world_size: int) -> None:
     batch = Batch.from_data_list([data], device=device) if rank == 0 else None
     local_batch = dd.partition(batch)
 
-    dd._prime_forces(local_batch)
+    dd._prime_forces(local_batch, dd._active_graph_mask(local_batch))
```
dallasfoster: This test is not sufficient to cover new code changes. Please consider adding a hook and testing, like so:

```python
seen = []

class _Probe:
    stage, frequency = DynamicsStage.AFTER_STEP, 1
    def __call__(self, ctx, stage):
        seen.append(ctx.active_graph_mask)

dd.register_hook(_Probe())
local_batch, _ = dd.step(local_batch)

mask = seen[-1]
assert mask is not None
assert mask.shape == (local_batch.num_graphs,)
assert bool(mask.all())            # nothing has graduated yet
```

dallasfoster: Worth adding a `HookScope.GLOBAL` probe too, since that branch builds its
context from `self._gather_all(batch)` rather than the local batch — a different
object, and the one place a shape mismatch could appear.

nvalchemi/dynamics/base.py
```diff
@@ -1667,12 +1680,26 @@ def __repr__(self) -> str:
             f"hooks={n_hooks})"
         )
 
-    def _call_hooks(self, stage: DynamicsStage, batch: Batch) -> None:
+    def _call_hooks(
+        self,
+        stage: DynamicsStage,
+        batch: Batch,
+        active_graph_mask: torch.Tensor | None,
```
dallasfoster: The DomainParallel I believe has active_graph_mask: torch.Tensor | None. I would suggest aligning on a single definition/contract. I even believe that HookRegistryMixin uses kwargs for capturing something like this.

test/distributed/model/test_multigpu.py
ys-teh: Updated the test in [e224792](https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/e224792a09274f2b96b886da8ad2f28aedf19c73).

Note that `active_graph_mask` is mostly redundant for the current domain parallel use case, where each rank owns a shard of the same system rather than a subset of multiple graphs. So `active_graph_mask` should be `None`. If domain parallel later supports ranks carrying shards from multiple graphs with mixed active statuses, that behavior should be covered already but it may still need additional testing.

nvalchemi/dynamics/base.py
ys-teh: Updated to initialize active_graph_mask = None. [e224792](https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/e224792a09274f2b96b886da8ad2f28aedf19c73)

nvalchemi/dynamics/base.py
```diff
@@ -2135,17 +2167,41 @@ def step(self, batch: Batch) -> tuple[Batch, torch.Tensor | None]:
                 elif val.shape[0] == batch.num_graphs:
                     saved[field] = val[sys_mask].clone()
 
+        # Prime once before the hook-wrapped step so the optimizer's first
+        # pre_update sees the same valid forces/stress as subsequent steps.
+        if not self._forces_primed:
+            self._prime_forces(batch, active_graph_mask)
```
laserkelvin, line 2178: `suggestion:` Priming runs the model once here, then the first step runs it again — and both fire the `BEFORE_COMPUTE`/`AFTER_COMPUTE` hooks at `step_count == 0`. A hook that watches compute (e.g. an energy logger) sees step 0 twice. Greptile flagged the same thing above; confirming it with a runnable check:

```python
import torch
from nvalchemi.dynamics.base import DynamicsStage
from nvalchemi.dynamics.demo import DemoDynamics

class Counter:
    stage = DynamicsStage.AFTER_COMPUTE
    frequency = 1
    def __init__(self):
        self.calls = []
    def __call__(self, ctx, stage):
        self.calls.append(ctx.step_count)

batch = Batch.from_data_list(
    [AtomicData(atomic_numbers=torch.tensor([6]), positions=torch.zeros(1, 3))]
)
batch.forces = torch.zeros(1, 3)
batch.energies = torch.zeros(1, 1)

dyn = DemoDynamics(model=DemoModelWrapper(DemoModel()), n_steps=2, dt=1.0)
hook = Counter()
dyn.register_hook(hook)
dyn.run(batch)

print(hook.calls)  # expect [0, 1] for two steps; got [0, 0, 1]
```

I assume the contract is one hook call per model compute, stamped with the step that compute belongs to. If the extra step-0 call is intended, one line in the `step()` docstring saying so would be enough; otherwise skip hook dispatch during priming.

nvalchemi/dynamics/base.py
```diff
@@ -3491,6 +3769,40 @@ def _close_hooks(self) -> None:
...
+        if status.dim() == 2:
+            status = status.squeeze(-1)
+        status = status[: batch.num_graphs]
+        for status_code, dynamics in self.sub_stages:
+            dynamics._call_hooks(
+                DynamicsStage.AFTER_COMPUTE,
+                batch,
+                status == status_code,
+            )
+
```
laserkelvin: `nit:` This file currently fails `ruff format --check` — trailing whitespace here, plus a few blank-line spots — and `docs/modules/dynamics/fused_stage.rst` has a trailing space in the new section. `make lint` (pre-commit ruff-format + trailing-whitespace) will go red on CI. One `make format` pass clears all of it.

laserkelvin (commented):
Two in-line comments: one on compute hooks firing twice at step 0 during force priming (with a small reproducer), one lint nit that will fail pre-commit CI.

nvalchemi/dynamics/base.py
ys-teh: Yes, I just updated the docstring to explain it. Previously I considered handling initial force priming as a dedicated iteration by masking the update phases and using that iteration’s normal compute, similar to transition repriming in the updated FusedStage. That would prevent the step count issue. But, that would change the existing first-step behavior in FusedStage and require a broader change, so I didn't do it.

ys-teh:
/ok to test [3a286fb](https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/3a286fb9140c6a4a745717dd0820ff3a1694ecb2)

laserkelvin:
/ok to test [6d3e187](https://github.com/NVIDIA/nvalchemi-toolkit/pull/181/commits/6d3e18703ef321fb4937002c54ea527d6b9f0aaf)

laserkelvin (approved):
TBH I don't *love* the fact that it launches twice, but I don't know if there's a more elegant (but correct) solution. Approving since everything is addressed.

PR #182, opened 2026-08-26
Fix/dynamics hooks and order

ys-teh:

## Description

Fix dynamics hook dispatch so hooks are no longer silently skipped and their
ordering is consistent across `BaseDynamics`, `FusedStage`, and fused
sub-stages. This also adds the one-time `ON_ADMISSION` stage, provides clear
warnings for unsupported fused-level stages, and deprecates the separate
`FusedStage.register_fused_hook()` API.

#181 needs to be merged before this.

Deferred for future PRs:
- ON_ADMISSION stage on DistributedPipeline / DomainParallel
- Efficient force priming on DistributedPipeline when using in-flight batching
- Better handling of optimizer / integrator states with hooks, e.g., FreezeAtomsHook, for correct behavior.

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds functionality)
- [ ] Performance improvement
- [X] Documentation update

## Changes Made

- Complete the hook lifecycle for `FusedStage`, including previously missing hook calls. Warn when a hook is registered on `FusedStage` for a sub-stage-only lifecycle
  boundary instead of silently accepting a hook that will never run.
- Add `DynamicsStage.ON_ADMISSION = -1` for validation, shape-dependent
  allocation, and other one-time batch setup before force priming. Admission
  remains outside the compiled fused step.
- Deprecate `FusedStage.fused_hooks` and `FusedStage.register_fused_hook()` in favor of the standard hook storage and `register_hook()`.
- Update the dynamics lifecycle documentation, diagrams, API guidance, and changelog.

## Testing

- `uv run --extra cu13 pytest test/dynamics/test_base.py test/dynamics/test_single_loop.py test/dynamics/test_bias_hook.py test/dynamics/test_cell_align.py test/dynamics/test_freeze_hook.py test/dynamics/test_periodic_hook.py test/dynamics/test_safety_hooks.py test/hooks/test_context.py`
  — 351 passed, 1 slow test skipped.
- `pytest test/dynamics test/hooks/test_neighbor_list_hook.py test/hooks/test_neighbor_list_nve_conservation.py test/hooks/test_stage_timing_hook.py` on 2 GPUs
  — 1109 passed, 2 skipped.
- [ ] Linting passes (`make lint`)
- [X] New tests added for new functionality meets coverage expectations?

## Checklist

- [X] I have read and understand the [Contributing Guidelines](../CONTRIBUTING.md)
- [X] I have updated the [CHANGELOG.md](../CHANGELOG.md)
- [X] I have performed a self-review of my code
- [X] I have added docstrings to new functions/classes
- [X] I have updated the documentation (if applicable)

## Additional Notes

Below are the rendered images in the updated documentation.

BaseDynamics workflow (`docs/modules/dynamics/hooks.rst` line 29):
image

FusedStage workflow (`docs/modules/dynamics/hooks.rst` line 341):
image

Overall workflow (`docs/modules/dynamics/overview.rst` line 27):
image

> [!TIP]

### DIFF
--- a/.claude/skills/nvalchemi-dynamics-hooks/SKILL.md
+++ b/.claude/skills/nvalchemi-dynamics-hooks/SKILL.md
@@ -77,6 +77,7 @@ class HookContext:
 class DynamicsContext(HookContext):
     step_count: int = 0
     converged_mask: torch.Tensor | None = None
+    active_graph_mask: torch.Tensor | None = None
 ```
 
 Access batch data via `ctx.batch` and dynamics step info via `ctx.step_count`.
@@ -87,9 +88,11 @@ Access batch data via `ctx.batch` and dynamics step info via `ctx.step_count`.
 
 ### Dynamics — `DynamicsStage`
 
-Each `step()` call fires hooks at 9 stages in this order:
+Dynamics exposes 10 lifecycle stages. `ON_ADMISSION` fires once when a
+batch is admitted, while the remaining 9 stages fire within each `step()`:
 
 ```text
+ON_ADMISSION (-1)  ← once before force priming and the first step
 BEFORE_STEP (0)
   BEFORE_PRE_UPDATE (1)  →  pre_update()  →  AFTER_PRE_UPDATE (2)
   BEFORE_COMPUTE (3)     →  compute()      →  AFTER_COMPUTE (4)
@@ -102,11 +105,24 @@ ON_CONVERGE (8)   ← only if convergence detected
 
 | Goal | Stage |
 |------|-------|
+| Validate or allocate for a newly admitted batch | `DynamicsStage.ON_ADMISSION` |
 | Modify forces/energy after model | `DynamicsStage.AFTER_COMPUTE` |
 | Observe final state (logging, snapshots) | `DynamicsStage.AFTER_STEP` |
 | Wrap positions after velocity update | `DynamicsStage.AFTER_POST_UPDATE` |
 | Instrument timing / profiling | `DynamicsStage.BEFORE_STEP` |
 | React to convergence | `DynamicsStage.ON_CONVERGE` |
+
+`ON_ADMISSION` is reset for every new `run()` and for managed membership
+changes such as refill or pipeline communication. In `FusedStage`, it runs
+outside compiled `_step_impl`, making it suitable for shape-dependent allocation
+and Python setup that per-step hooks cannot safely perform under `fullgraph=True`.
+
+In `FusedStage`, fused-level hooks wrap sub-stage hooks at every shared boundary:
+fused `BEFORE_*` hooks run before the corresponding sub-stage loop, and fused
+`AFTER_*` hooks run after it. This includes the pre-update and post-update
+boundaries. Fused hooks receive the overall active mask, while hooks registered
+on a sub-stage receive only that sub-stage's status mask. `ON_CONVERGE` remains
+sub-stage-only because convergence is evaluated independently per sub-stage.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Unreleased
 
+### Fixed
+
+- **Dynamics hook lifecycle** — add `ON_ADMISSION` for one-time batch setup
+  before force priming and outside compiled fused steps. Fused stages now fire
+  shared step and compute hooks at both fused and sub-stage levels with
+  consistent nesting.
+
+### Deprecated
+
+- `FusedStage.register_fused_hook()`. Use the inherited `register_hook()`
+  method instead; hooks on a `FusedStage` already observe the complete fused
+  batch.
+
+## 0.2.0 — 2026-08-07
+
 ### Added
 
 - Domain decomposition for distributed inference and dynamics: a spatial halo
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,22 @@
 
 ## Unreleased
 
+### Added
+
+- `FusedStage(reprime_on_entry=...)` — status codes whose newly entering
+  graphs skip one integrator update so the shared compute and target-stage
+  `AFTER_COMPUTE` hooks can refresh forces under the new stage's context
+  before it advances them.
+
 ### Fixed
 
 - **Dynamics hook lifecycle** — add `ON_ADMISSION` for one-time batch setup
   before force priming and outside compiled fused steps. Fused stages now fire
   shared step and compute hooks at both fused and sub-stage levels with
   consistent nesting.
+- **`FusedStage` force priming** — a dynamics instance's own adaptive
+  optimizer state (e.g. FIRE's per-graph `dt`/`alpha`/step counters in
+  `self._state`) is now preserved across masked `pre_update`/`post_update` calls.
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Added
 
+- Add support for PEFT fine-tuning within `FineTuningStrategy`, including
+  LoRA workflows with `LoRAConfig`, `load_peft_checkpoint_into_model`,
+  and base-model fingerprint checks for PEFT checkpoint loading.
 - `FusedStage(reprime_on_entry=...)` — status codes whose newly entering
   graphs skip one integrator update so the shared compute and target-stage
   `AFTER_COMPUTE` hooks can refresh forces under the new stage's context
--- a/docs/modules/dynamics/hooks.rst
+++ b/docs/modules/dynamics/hooks.rst
@@ -258,14 +258,21 @@ FreezeAtomsHook
 ...............
 
 :class:`~nvalchemi.dynamics.hooks.FreezeAtomsHook` keeps selected atoms fixed:
-it snapshots positions and clears frozen state at ``BEFORE_PRE_UPDATE``, restores
-the constrained geometry at ``AFTER_PRE_UPDATE``, clears model forces before the
-post-update, and restores positions with zeroed velocities at
-``AFTER_POST_UPDATE``.
+it fires at five stages. At ``BEFORE_PRE_UPDATE`` it snapshots positions and
+clears prior forces and velocities; at ``AFTER_PRE_UPDATE`` it restores the
+constrained geometry before compute preparation; at ``AFTER_COMPUTE`` it clears
+new model forces when ``zero_forces=True``; at ``BEFORE_POST_UPDATE`` it always
+clears frozen-atom forces; and at ``AFTER_POST_UPDATE`` it restores positions
+and zeroes velocities.
 
-``categories`` is a string or list of strings matching atom type categories in
-the batch (for example, ``"substrate"`` or ``["substrate", "boundary"]``). Only
-atoms in the listed categories are frozen; all others evolve freely.
+``freeze_category`` is the integer ``batch.atom_categories`` value that marks
+frozen atoms. It defaults to :attr:`~nvalchemi._typing.AtomCategory.SPECIAL`.
+Only atoms matching that value are frozen; all others evolve freely.
+
+Set ``zero_forces=False`` to expose raw frozen-atom model forces to
+``AFTER_COMPUTE`` observers. Those forces are still cleared at
+``BEFORE_POST_UPDATE``, before the second integrator update, so they cannot
+move the frozen atoms.
 
 Use this hook for partial-system relaxations (freeze the substrate, relax the
 adsorbate), slab calculations (freeze bottom layers), or any configuration
--- a/docs/modules/dynamics/hooks.rst
+++ b/docs/modules/dynamics/hooks.rst
@@ -312,23 +312,17 @@ checking so the detector sees the corrected forces:
 Hooks inside ``FusedStage``
 ---------------------------
 
-When hooks are registered on sub-stage dynamics inside a
-:class:`~nvalchemi.dynamics.FusedStage`, their firing semantics differ
-slightly from standalone execution:
+Hooks may be registered directly on a
+:class:`~nvalchemi.dynamics.FusedStage` or on any of its sub-stages. Shared step
+and compute boundaries fire at both levels. Integrator update boundaries fire
+only on sub-stages, which own the corresponding masked updates. Fused-stage
+hooks receive the full batch with the overall active mask, while each sub-stage
+hook receives the same batch with its status-specific mask.
 
-**Fired on each sub-stage:**
-
-- ``BEFORE_STEP``, ``AFTER_COMPUTE``, ``BEFORE_PRE_UPDATE``,
-  ``AFTER_POST_UPDATE``, ``AFTER_STEP``, ``ON_CONVERGE``
-
-**Not fired on sub-stages** (because the forward pass is shared):
-
-- ``BEFORE_COMPUTE``, ``AFTER_PRE_UPDATE``, ``BEFORE_POST_UPDATE``
-
-This means safety hooks (``NaNDetectorHook``, ``MaxForceClampHook``)
-and observer hooks (``LoggingHook``, ``SnapshotHook``) work as expected
-inside fused stages, since they fire at ``AFTER_COMPUTE`` or
-``AFTER_STEP``.
+At shared ``BEFORE_*`` boundaries, the fused-stage hooks fire before the
+sub-stage hooks. At shared ``AFTER_*`` boundaries, the sub-stage hooks fire
+before the fused-stage hooks. The update boundaries and ``ON_CONVERGE`` are
+sub-stage-only; registering those hooks on ``FusedStage`` emits a warning.
 
 Hook ordering inside a fused step:
 
@@ -341,46 +335,41 @@ Hook ordering inside a fused step:
        node [fontsize=11 shape=box style="rounded,filled" fillcolor="#1a1a1a"]
        edge [fontsize=10 style=bold]
 
-       subgraph cluster_before {
-           label="for each sub-stage"
-           style=dashed
-           color="#76b900"
-           fontcolor="#76b900"
-           fontsize=10
-           BEFORE_STEP [label="BEFORE_STEP hooks"]
-       }
+       fused_before_step [label="FusedStage BEFORE_STEP hooks" fillcolor="#4a3315"]
+       sub_before_step [label="each sub-stage BEFORE_STEP hooks\n(in sub-stage order)"]
 
-       compute [label="single compute()" fillcolor="#4a3315"]
-
-       subgraph cluster_after_compute {
-           label="for each sub-stage"
-           style=dashed
-           color="#76b900"
-           fontcolor="#76b900"
-           fontsize=10
-           AFTER_COMPUTE [label="AFTER_COMPUTE hooks"]
-       }
-
-       subgraph cluster_update {
+       subgraph cluster_pre_update {
            label="for each sub-stage"
            style=dashed
            color="#76b900"
            fontcolor="#76b900"
            fontsize=10
            BEFORE_PRE [label="BEFORE_PRE_UPDATE hooks"]
-           masked     [label="masked_update()\n(if samples match status)" fillcolor="#1a1a1a"]
-           AFTER_POST [label="AFTER_POST_UPDATE hooks"]
-           BEFORE_PRE -> masked -> AFTER_POST
+           pre_update [label="masked_pre_update()" fillcolor="#1a1a1a"]
+           AFTER_PRE [label="AFTER_PRE_UPDATE hooks"]
+           BEFORE_PRE -> pre_update -> AFTER_PRE
        }
 
-       subgraph cluster_after_step {
+       fused_before_compute [label="FusedStage BEFORE_COMPUTE hooks" fillcolor="#4a3315"]
+       sub_before_compute [label="each sub-stage BEFORE_COMPUTE hooks\n(in sub-stage order)"]
+       compute [label="single shared compute()" fillcolor="#4a3315"]
+       sub_after_compute [label="each sub-stage AFTER_COMPUTE hooks\n(in sub-stage order)"]
+       fused_after_compute [label="FusedStage AFTER_COMPUTE hooks" fillcolor="#4a3315"]
+
+       subgraph cluster_post_update {
            label="for each sub-stage"
            style=dashed
            color="#76b900"
            fontcolor="#76b900"
            fontsize=10
-           AFTER_STEP [label="AFTER_STEP hooks"]
+           BEFORE_POST [label="BEFORE_POST_UPDATE hooks"]
+           post_update [label="masked_post_update()" fillcolor="#1a1a1a"]
+           AFTER_POST [label="AFTER_POST_UPDATE hooks"]
+           BEFORE_POST -> post_update -> AFTER_POST
        }
+
+       sub_after_step [label="each sub-stage AFTER_STEP hooks\n(in sub-stage order)"]
+       fused_after_step [label="FusedStage AFTER_STEP hooks" fillcolor="#4a3315"]
 
        subgraph cluster_converge {
            label="for each sub-stage"
@@ -388,18 +377,29 @@ Hook ordering inside a fused step:
            color="#76b900"
            fontcolor="#76b900"
            fontsize=10
-           conv_check  [label="convergence check" fillcolor="#1a1a1a"]
-           ON_CONVERGE [label="ON_CONVERGE hooks" fillcolor="#4a3315"]
-           conv_check -> ON_CONVERGE [style=dashed label="if converged"]
+           conv_check  [label="convergence evaluation" fillcolor="#1a1a1a"]
+           ON_CONVERGE [label="ON_CONVERGE hooks\n(with convergence mask)"]
+           conv_check -> ON_CONVERGE [style=dashed]
        }
 
-       BEFORE_STEP -> compute
-       compute -> AFTER_COMPUTE
-       AFTER_COMPUTE -> BEFORE_PRE
-       AFTER_POST -> AFTER_STEP
-       AFTER_STEP -> conv_check
+       fused_before_step -> sub_before_step
+       sub_before_step -> BEFORE_PRE [lhead=cluster_pre_update]
+       AFTER_PRE -> fused_before_compute [ltail=cluster_pre_update]
+       fused_before_compute -> sub_before_compute
+       sub_before_compute -> compute
+       compute -> sub_after_compute
+       sub_after_compute -> fused_after_compute
+       fused_after_compute -> BEFORE_POST [lhead=cluster_post_update]
+       AFTER_POST -> sub_after_step [ltail=cluster_post_update]
+       sub_after_step -> fused_after_step
+       fused_after_step -> conv_check [lhead=cluster_converge]
    }
 
+Initial force priming follows the same nested ``BEFORE_COMPUTE`` and
+``AFTER_COMPUTE`` ordering. Safety hooks (``NaNDetectorHook``,
+``MaxForceClampHook``) and observer hooks (``LoggingHook``, ``SnapshotHook``)
+therefore behave consistently whether they are registered on the fused stage
+or on a specific sub-stage.
 
 API reference
 -------------
--- a/docs/modules/dynamics/hooks.rst
+++ b/docs/modules/dynamics/hooks.rst
@@ -22,11 +22,12 @@ For the general hook protocol, context, and registry see
 DynamicsStage
 --------------
 
-:class:`~nvalchemi.dynamics.base.DynamicsStage` enumerates the nine
-hook-firing points within a single dynamics step:
+:class:`~nvalchemi.dynamics.base.DynamicsStage` enumerates ten lifecycle
+hook-firing points: ``ON_ADMISSION`` for batch setup, followed by nine stages
+within each dynamics step:
 
 .. graphviz::
-   :caption: DynamicsStage hook firing points within a single step.
+   :caption: DynamicsStage lifecycle hook firing points.
 
    digraph dynamics_stages {
        rankdir=TB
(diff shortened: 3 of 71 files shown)


nvalchemi/distributed/domain_parallel.py
```diff
@@ -262,13 +262,18 @@ def step(self, batch: Batch) -> tuple[Batch, torch.Tensor | None]:
...
+            self._prime_forces(batch, active_graph_mask)
             self._forces_primed = True
 
         # 1. Outer BEFORE_STEP hooks.
-        self._call_hooks(DynamicsStage.BEFORE_STEP, batch)
+        self._call_hooks(
+            DynamicsStage.BEFORE_STEP,
+            batch,
+            active_graph_mask,
+        )
```
dallasfoster, line 278: `_ensure_admission_initialized` is the only thing that dispatches the new stage,
and it is called from just two places — `BaseDynamics.step` and the `FusedStage`
step. `DomainParallel` overrides `step` and calls neither; `ON_ADMISSION` does
not appear anywhere in `domain_parallel.py`:

I think the fix is to call `self._ensure_admission_initialized(batch)` at the top
of `DomainParallel.step`, before the `_dist_model is None` fallback, so the
wrapper fires admission on its own registry once per system on either branch.

While you are there: `DomainParallel.__init__` forwards only the model, so
`self.exit_status` defaults to `1` and is never taken from the wrapped dynamics,
while `_active_graph_mask` (`domain_parallel.py:681`) reads
`self._dynamics.exit_status`.

test/distributed/model/test_multigpu.py
dallasfoster, line 1: `ON_ADMISSION` is covered in `test_single_loop.py` and `test_base.py`, but the
three test files this PR touches carry zero `multigpu` marks, so none of the new
stage or reordering runs on the 2-GPU job.

`domain_parallel.py` never mentions
`ON_ADMISSION`, but it inherits the dispatch through its own `_call_hooks`
override — so under DD the new stage fires on a DP-specific path that computes
its mask differently and has no test at any world size.

Could we add one multigpu test that registers a counting hook on a
`DomainParallel`, runs a couple of steps at `world_size=2`, and asserts
`ON_ADMISSION` fired once per system with the expected maskT

This comment is similar to the hole in #181 that opens from the other side, so one test settles both.

nvalchemi/distributed/domain_parallel.py
ys-teh: Thanks, it should now be resolved in [f75d7b1](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/f75d7b18da1b8660ff40c20b45276342a8ec37da). I added `self._ensure_admission_initialized(batch)` after the `_dist_model is None` fallback instead since there will be readmission with dynamics.run anyway.

test/distributed/model/test_multigpu.py
ys-teh: Updated in [f75d7b1](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/f75d7b18da1b8660ff40c20b45276342a8ec37da).

ys-teh:
/ok to test [2dfe578](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/2dfe5782f908b964bbb11ce6352caa59884d1af5)

nvalchemi/dynamics/base.py
```diff
@@ -3460,37 +3857,80 @@ def all_complete(batch: Batch, exit_status: int) -> bool:
...
-            if hook_id in seen:
-                continue
-            seen.add(hook_id)
-            if hasattr(hook, "__exit__"):
-                hook.__exit__(None, None, None)
-            elif hasattr(hook, "close"):
-                hook.close()
-
         for _, dynamics in self.sub_stages:
             dynamics._close_hooks()
```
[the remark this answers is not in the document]
ys-teh: super()._open_hooks() and super()._close_hooks() manage the fused stage’s inherited self.hooks. Duplicate lifecycle calls occur only if a caller explicitly registers the same hook instance at multiple hierarchy levels. This is pre-existing behavior and is unrelated to replacing fused_hooks with the inherited registry.

.claude/skills/nvalchemi-dynamics-hooks/SKILL.md
```diff
@@ -102,12 +105,25 @@ ON_CONVERGE (8)   ← only if convergence detected
...
 | Wrap positions after velocity update | `DynamicsStage.AFTER_POST_UPDATE` |
 | Instrument timing / profiling | `DynamicsStage.BEFORE_STEP` |
 | React to convergence | `DynamicsStage.ON_CONVERGE` |
 
+`ON_ADMISSION` is reset for every new `run()` and for managed membership
+changes such as refill or pipeline communication. In `FusedStage`, it runs
+outside compiled `_step_impl`, making it suitable for shape-dependent allocation
+and Python setup that per-step hooks cannot safely perform under `fullgraph=True`.
+
+In `FusedStage`, fused-level hooks wrap sub-stage hooks at every shared boundary:
```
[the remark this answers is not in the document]
ys-teh: Addressed in [92bbf3b](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/92bbf3bf4cca4dcbf4812e15773160d8e935a5bf)

laserkelvin:
/ok to test [92bbf3b](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/92bbf3bf4cca4dcbf4812e15773160d8e935a5bf)

nvalchemi/dynamics/base.py
```diff
@@ -3513,26 +3626,34 @@ def _step_impl(self, batch: Batch) -> tuple[Batch, torch.Tensor | None]:
...
         self._call_hooks(
             DynamicsStage.BEFORE_COMPUTE,
             batch,
             overall_active_graph_mask,
         )
 
+        for (_, dynamics), active_graph_mask in zip(
+            self.sub_stages, stage_active_masks, strict=True
+        ):
+            dynamics._call_hooks(
```
laserkelvin, line 3631: `suggestion:` This dispatch (sub-stage `BEFORE_COMPUTE` inside `_step_impl`) is the right fix — these hooks were silently dead before — but one heads-up on the interaction with the documented `compile(fullgraph=True)` config: `_step_impl` is the `torch.compile` target, and the canonical `BEFORE_COMPUTE` hook, `NeighborListHook`, calls `@torch.compiler.disable`d helpers, so the first in-step firing fails compilation. Minimal reproducer (CPU):

```python
import torch
from nvalchemi.dynamics.base import BaseDynamics, DynamicsStage, FusedStage
from nvalchemi.hooks.neighbor_list import NeighborConfig, NeighborListHook

batch = Batch.from_data_list(
    [AtomicData(atomic_numbers=torch.tensor([6, 6]), positions=torch.randn(2, 3))]
)
batch.forces = torch.zeros(batch.num_nodes, 3)
batch.energy = torch.zeros(batch.num_graphs, 1)
batch.status = torch.zeros(batch.num_graphs, 1, dtype=torch.long)

sub = BaseDynamics(model=DemoModelWrapper(DemoModel()))
sub.register_hook(
    NeighborListHook(NeighborConfig(cutoff=5.0), stage=DynamicsStage.BEFORE_COMPUTE)
)
fused = FusedStage(
    sub_stages=[(0, sub)],
    compile_step=True,
    compile_kwargs={"backend": "eager", "fullgraph": True},
    device_type="cpu",
)
fused.step(batch)
# torch._dynamo.exc.Unsupported: Skip calling torch.compiler.disable()'d
# function `_write_neighbor_data_to_batch`
```

Under the default `fullgraph=False` it runs but graph-splits around the hook every step. Not asking for compile-safe hooks in this PR — a short docs note ("in-step hooks run inside the compiled step and must be compile-safe; `NeighborListHook` is not today") would save users a confusing error, and compile-safety for stock hooks can be follow-up work.

nvalchemi/dynamics/base.py
```diff
@@ -4553,6 +4704,8 @@ def step(self) -> None:
                 if stage.active_batch is not None:
                     if stage.active_batch.device != stage.device:
                         stage.active_batch = stage.active_batch.to(stage.device)
+                    stage._admission_initialized = False
```
laserkelvin, line 4694: `question:` This initial-build branch resets both one-time flags on a fresh batch, but the rebuild-from-sampler branch below (`base.py:4770`, when `active_batch is None` after the step) resets only `_forces_primed`. If a stage can reach that branch with `_admission_initialized` still `True`, `ON_ADMISSION` never fires for the rebuilt batch. From my reading, all of today's paths into it reset admission earlier (`_remove_converged_final_stage`, `refill_check`), so this may be moot — but should the rebuild branch reset both for symmetry, so the invariant doesn't depend on which distant method happened to run first?

laserkelvin (commented):
Overall this is a solid fix — the dispatch semantics are consistent across `BaseDynamics`/`FusedStage`/sub-stages now, the ordering tests are good, and the docs match the implementation. Left a handful of in-line comments: mostly questions and nits, plus two discussion points (the stage-counter cascade at `base.py:3738`, which is pre-existing on main and not a blocker here, and the fused-level update-boundary masks).

One suggestion for a file not in this diff: `HookRegistryMixin._call_hooks` (`nvalchemi/hooks/_registry.py`) always builds the hook context before checking whether any hook is registered. An `if not self.hooks: return` early return is provably safe (hooks still fire; compiled-with-hook behavior byte-identical) and cuts dynamo trace time ~3x in a microbenchmark (3.65s -> 1.82s cold compile, 2 sub-stages) because dynamo stops inlining the per-dispatch context builds. Eager saves ~1us/dispatch. Compiled per-step runtime is unchanged either way — dynamo already erases the no-op dispatch (0 graph breaks). Optional polish only.

nvalchemi/dynamics/base.py
```diff
@@ -3795,17 +3913,44 @@ def _prime_forces(
         active_graph_mask : torch.Tensor | None
             Boolean mask selecting all active graphs.
         """
-        super()._prime_forces(batch, active_graph_mask)
-
+        self._ensure_state_initialized(batch)
```
laserkelvin, line 3903: `issue:` The rewritten `_prime_forces` runs the full nested compute-hook lifecycle (fused `BEFORE_COMPUTE` -> sub-stage `BEFORE_COMPUTE` in order -> `compute()` -> sub-stage `AFTER_COMPUTE` -> fused `AFTER_COMPUTE`), but no test observes it: the ordering tests all set `fused._forces_primed = True` before `step()` (e.g. `test_single_loop.py:1972`), and the DomainParallel priming test patches `_prime_forces` to raise. A regression that drops or misnests a priming dispatch (say, a neighbor-list hook missing at step 0) would fail no test. Could you add a unit test pinning the priming order? Concretely, with priming enabled and ordered recording hooks at `BEFORE_COMPUTE`/`AFTER_COMPUTE` on the fused stage plus two sub-stages (the existing `_OrderedHook` pattern works): assert (1) the priming bracket fires before the first `BEFORE_STEP`, ordered fused -> sub0 -> sub1 for `BEFORE_COMPUTE` and sub0 -> sub1 -> fused for `AFTER_COMPUTE`; (2) the fused hook receives the overall active mask (`status < exit_status`) and each sub-stage hook receives `status == status_code`; (3) `compute()` runs exactly once during priming.

nvalchemi/dynamics/base.py
```diff
@@ -3496,6 +3604,11 @@ def _step_impl(self, batch: Batch) -> tuple[Batch, torch.Tensor | None]:
         # Phase 1 — pre_update for each sub-stage.
         # This moves positions to r(t+dt) so that the shared compute can
         # evaluate forces at the correct (updated) positions.
+        self._call_hooks(
```
laserkelvin, line 3594: `question:` The four fused-level update-boundary dispatches (here and the matching `AFTER_PRE_UPDATE`/`BEFORE_POST_UPDATE`/`AFTER_POST_UPDATE` below) receive `overall_active_graph_mask`, which **includes** `reprime_pending` graphs, while the sub-stage dispatches at the same boundaries receive `update_graph_mask`, which excludes them. So a fused-level *mutating* hook would touch a graph that is sitting the step out. Minimal reproducer:

```python
fused = FusedStage(sub_stages=[(0, sub0), (1, sub1)],
                   exit_status=2, reprime_on_entry={1})
# sub0 has n_steps=1; hooks at BEFORE_PRE_UPDATE on fused, sub0, sub1
# capture ctx.active_graph_mask. One graph, status 0.

fused.step(batch)  # graph updated in stage 0, migrates to 1, marked reprime-pending
fused.step(batch)  # observed masks at BEFORE_PRE_UPDATE, step_count=1:
#   fused: [True]   <- includes the reprime-pending graph
#   sub1:  [False]  <- correctly excludes it at the same boundary
```

If the broad fused mask is intended (whole-batch observation), could you say so — the comment at `base.py:3578` ("masked out of the ... hook and integrator phases") predates these dispatches and no longer matches, and `test_fused_update_hooks_receive_overall_active_mask` pins the broad mask but uses no reprime-pending graphs, so it can't distinguish the two readings. A parametrized case with `reprime_on_entry` set would pin it.

nvalchemi/dynamics/base.py
```diff
@@ -1694,6 +1710,29 @@ def _call_hooks(
...
+        if self._admission_initialized:
+            return
+        status = getattr(batch, "status", None)
+        if status is None:
+            active_graph_mask = None
+        else:
+            status = status.squeeze(-1) if status.dim() == 2 else status
+            active_graph_mask = status[: batch.num_graphs] < self.exit_status
+        self._call_hooks(
+            DynamicsStage.ON_ADMISSION,
```
laserkelvin, line 1727: `question:` Admission dispatch goes through the normal frequency gate (`step_count % hook.frequency == 0`), but `step_count` is never reset across `run()` or refill — so an `ON_ADMISSION` hook with `frequency > 1` can silently skip an admission (e.g. a refill landing at step 7 with `frequency=2`). `frequency > 1` on a once-per-event stage is odd input, so this may be a non-issue in practice — is the gating intended here? If yes, a one-line docs note would do; if not, bypass the gate for `ON_ADMISSION` or warn at `register_hook` time.

CHANGELOG.md
```diff
@@ -7,6 +7,31 @@
...
   LoRA workflows with `LoRAConfig`, `load_peft_checkpoint_into_model`,
   and base-model fingerprint checks for PEFT checkpoint loading.
+- `FusedStage(reprime_on_entry=...)` — status codes whose newly entering
+  graphs skip one integrator update so the shared compute and target-stage
+  `AFTER_COMPUTE` hooks can refresh forces under the new stage's context
+  before it advances them.
+
+### Fixed
+
+- **Dynamics hook lifecycle** — add `ON_ADMISSION` for one-time batch setup
```
laserkelvin: `suggestion:` Small changelog alignment — this entry is filed under "Fixed", but `ON_ADMISSION` is a new stage, so users scanning "Added" will miss it. Also worth a short behavior-change note for existing users: fused-level hooks at the four update boundaries and sub-stage `BEFORE_COMPUTE` never fired before and now do; fused-level `AFTER_COMPUTE` moved from before the sub-stage loop to after it; and anyone who worked around the old bug by registering the same hook at both levels will now see it fire twice per boundary.

test/dynamics/test_single_loop.py
```diff
@@ -1867,6 +1940,83 @@ def setup_method(self) -> None:
...
+            DynamicsStage.BEFORE_PRE_UPDATE,
+            DynamicsStage.AFTER_PRE_UPDATE,
+            DynamicsStage.BEFORE_COMPUTE,
+            DynamicsStage.AFTER_COMPUTE,
+            DynamicsStage.BEFORE_POST_UPDATE,
+            DynamicsStage.AFTER_POST_UPDATE,
+            DynamicsStage.AFTER_STEP,
+        ],
+    )
+    def test_fused_hooks_bracket_substage_hooks(self, stage: DynamicsStage) -> None:
```
laserkelvin, line 1963: `suggestion:` While adding ordering tests, four nearby gaps would round out coverage: (1) no public-path admission-reset test — the only refire test sets `dynamics._admission_initialized = False` by hand instead of driving a second `run()` or a refill; (2) admission mask contents are never asserted (fused `status < exit_status`, sub-stage `status == code`); (3) no `pytest.warns` covers the `register_fused_hook` DeprecationWarning or the ON_CONVERGE-on-FusedStage UserWarning; (4) `_restore_unmasked_state`'s cardinality `RuntimeError` branch is exercised nowhere.

nvalchemi/dynamics/base.py
```diff
@@ -3349,61 +3399,115 @@ def _ensure_bookkeeping_fields(self, batch: Batch) -> None:
...
-        ctx = self._build_context(
+        warnings.warn(
+            "FusedStage.register_fused_hook() is deprecated; use "
+            "FusedStage.register_hook() instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        self.register_hook(hook)
+
+    def _mark_reprime_entries(
```
laserkelvin, line 3472: `nit:` `_mark_reprime_entries` is defined twice in `FusedStage` — here and again at ~line 3512, with identical bodies. The second definition silently shadows this one, so this copy is dead code. Delete one.

nvalchemi/dynamics/base.py
```diff
@@ -1023,6 +1027,10 @@ def _buffer_to_batch(self, incoming_batch: Batch) -> None:
         else:
             combined = self.active_batch.to_data_list() + admitted
         self.active_batch = Batch.from_data_list(combined, device=incoming_batch.device)
+        if hasattr(self, "_admission_initialized"):
```
laserkelvin: `nit:` This guarded reset block repeats verbatim three times in `_CommunicationMixin` (also ~1126 and ~1330). Consider initializing both flags in `_CommunicationMixin.__init__` instead — then the `hasattr` guards can go away entirely.

nvalchemi/distributed/domain_parallel.py
```diff
@@ -87,6 +87,7 @@ def __init__(
         config: DomainConfig,
         **kwargs: Any,
     ) -> None:
+        kwargs["exit_status"] = dynamics.exit_status
```
laserkelvin, line 100: `question:` Unifying the thresholds is right, but an explicitly-passed `exit_status=` is silently discarded here (the new test pins `99` -> `3`). Should we warn when the caller's value differs from the wrapped dynamics', so the override isn't silent?

nvalchemi/dynamics/base.py
```diff
@@ -1834,6 +1873,39 @@ def _validate_batch_keys(self, batch: Batch) -> None:
     # Per-system integrator state management
     # ------------------------------------------------------------------
 
+    def _save_state_fields(self) -> dict[str, torch.Tensor]:
+        """Clone tensor-valued integrator state, preserving each field's shape."""
```
laserkelvin: `nit:` Docstring drift in three spots: (1) here — says "tensor-valued", but the `isinstance` filter was dropped and every field is cloned now (fine for in-tree state, but an out-of-tree dynamics with a non-tensor state field gets `AttributeError` instead of a skip; either restore the filter or reword); (2) `base.py:1991` still cites `masked_update` in the subclassing contract, though `FusedStage` uses `_masked_pre_update`/`_masked_post_update`; (3) `BaseDynamics.step()`'s docstring lists its phases starting at force priming and omits the admission dispatch its body performs at ~line 2219.

docs/modules/dynamics/hooks.rst
```diff
@@ -312,23 +330,33 @@ checking so the detector sees the corrected forces:
...
 
-When hooks are registered on sub-stage dynamics inside a
-:class:`~nvalchemi.dynamics.FusedStage`, their firing semantics differ
-slightly from standalone execution:
+Hooks may be registered directly on a
+:class:`~nvalchemi.dynamics.FusedStage` or on any of its sub-stages. Step,
+compute, and integrator update boundaries all fire at both levels. Fused-stage
+hooks receive the full batch with the overall active mask (all systems whose
+status is below ``exit_status``), while each sub-stage hook receives the same
+batch with its status-specific mask. At admission,
```
laserkelvin: `nit:` Two small accuracy issues in this file: (1) here — "each sub-stage hook receives the same batch with its status-specific mask" holds at step/compute boundaries, but at the four update boundaries sub-stage hooks get the narrower update-eligibility mask (excludes `reprime_pending`); the hooks skill got this distinction in 92bbf3b but the rst didn't. (2) The stage table (~line 112) still says `ON_CONVERGE` fires "only when the convergence hook detects converged samples" — on fused sub-stages, registered hooks fire every step and must consult `ctx.converged_mask` themselves.

docs/userguide/hooks.md
```diff
@@ -96,6 +96,10 @@ Dynamics engines pass {py:class}`~nvalchemi.hooks.DynamicsContext`, which adds:
 |-------|------|---------|
 | `step_count` | `int` | Current dynamics step |
 | `converged_mask` | `torch.Tensor \| None` | Samples that converged at the current hook stage |
+| `active_graph_mask` | `torch.Tensor \| None` | Systems active for the current fused or sub-stage dispatch |
+
+Hooks should respect `active_graph_mask` when set to preserve
```
laserkelvin: `suggestion:` "Hooks should respect `active_graph_mask`" needs a bit more for end-users to act on — what respecting means and when it matters. The built-in mutating hooks show the idiom: `sel = my_selection & ctx.active_graph_mask[batch.batch_idx]` at node level (`BiasedPotentialHook`, `FreezeAtomsHook`) or `& ctx.active_graph_mask` at graph level (`WrapPeriodicHook`). Suggested rewrite: mutating hooks should AND their selection with the mask — otherwise a sub-stage hook can mutate graphs owned by another stage or sitting out a reprime step — while read-only observation hooks may deliberately read the full batch (the Part 4 `StatusSnapshotHook` in the multistage example is such a case; reading everything is its purpose).

nvalchemi/dynamics/base.py
laserkelvin, line 1: `question:` (anchored file-level because the counter loop is unchanged context outside this PR's diff hunks; the line in question is `base.py:3738`. Pre-existing, not introduced by this PR — posting for visibility since the Greptile P1 thread above is open.) The counter advance keys on the live `migration_status`, so a graph that migrates into a stage *during this step* immediately counts that step toward the new stage's `n_steps` budget without receiving any of that stage's updates. With consecutive `n_steps=1` stages, a graph passes through a stage with zero updates applied (reproduced locally: 2 sub-stages, `n_steps=1` each, 1 graph — the stage-1 update hook saw 0 updates for a graph that graduated through it in a single step). This seems to contradict both the docs ("On the next step, it is processed by sub-stage 1"; the `reprime_on_entry` section's "entering graphs skip updates and counters") and the comment above ("eligible for a dynamics update *at the start of this iteration*", which reads like `stage_update_masks[i]`). Intended? If not, the fix is a one-liner and I'm happy to file it as a follow-up issue so it doesn't block this PR.

nvalchemi/dynamics/base.py
ys-teh: Updated in [cfb052e](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/cfb052e84335b69c3f4e02580b60c30134a8aa93) and also highlighted the nan detector hook.

nvalchemi/dynamics/base.py
ys-teh: Good catch. I’ve updated it to reset _admission_initialized alongside _forces_primed and extended the fused and base tests accordingly. [cfb052e](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/cfb052e84335b69c3f4e02580b60c30134a8aa93)

nvalchemi/dynamics/base.py
ys-teh: Added `test_force_priming_brackets_compute_hooks_before_first_step` in `test_single_loop.py`. [cfb052e](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/cfb052e84335b69c3f4e02580b60c30134a8aa93)

nvalchemi/dynamics/base.py
ys-teh: Thanks for flagging this. The previous behavior was incorrect. It's now addressed. [cfb052e](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/cfb052e84335b69c3f4e02580b60c30134a8aa93)

nvalchemi/dynamics/base.py
ys-teh: Deleted in [d369226](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/d369226e737dcbbdaa1b7d0ed6c7b0e2c636b7c6).

nvalchemi/dynamics/base.py
ys-teh: I think adding `ignore_frequency=True` for `ON_ADMISSION` would be the correct behavior. I did that and it involves changes to 10 files. [19ad439](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/19ad439fcdc07af38749054e440022880ae358b8)

test/dynamics/test_single_loop.py
ys-teh: Thanks for identifying these gaps. Addressed in [`f0aeed7`](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/f0aeed7ab956e16462115a06c11f376a6edffe07) using the existing nearby coverage:

1. Public-path admission reset is covered by `test_on_admission_ignores_frequency_across_runs` in `test_base.py` and `test_admission_hooks_ignore_frequency_across_runs` in `test_single_loop.py`; both call `run()` twice.
2. Extended the latter test to assert the fused and per-sub-stage admission masks across both runs.
3. Added `test_fused_hook_registration_warnings` in `test_single_loop.py` for both requested warnings.
4. Extended `test_masked_updates_preserve_inactive_sub_stage_state` in `test_state_management.py` to exercise the `_restore_unmasked_state()` cardinality error.

The affected test classes pass: 33 tests.

nvalchemi/dynamics/base.py
ys-teh: Addressed in [f0aeed7](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/f0aeed7ab956e16462115a06c11f376a6edffe07).

nvalchemi/distributed/domain_parallel.py
ys-teh: Added warning in [b898676](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/b8986763c61fb97e830496a31a11be5f47ac9606).

CHANGELOG.md
ys-teh: Addressed in [5e6bf53](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/5e6bf53226dadef53be1cee1347b210024efd566).

docs/modules/dynamics/hooks.rst
ys-teh: Addressed in [cfb052e8](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/cfb052e84335b69c3f4e02580b60c30134a8aa93) and [5e6bf532](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/5e6bf53226dadef53be1cee1347b210024efd566): clarified the update masks and fused sub-stage `ON_CONVERGE` behavior.

docs/userguide/hooks.md
ys-teh: Addressed in [5e6bf532](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/5e6bf53226dadef53be1cee1347b210024efd566).

nvalchemi/dynamics/base.py
ys-teh: Fixed in [e35c6c21](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/e35c6c21595b78a0f1b50d21b78bfb84f95ba348) by using `stage_update_masks[i]` instead of `overall_update_graph_mask`.

nvalchemi/dynamics/base.py
ys-teh: Addressed all these in [5e6bf532](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/5e6bf53226dadef53be1cee1347b210024efd566).

ys-teh:
/ok to test [49bf957](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/49bf9570ab8600734faad77c82dcdf2f48334621)

laserkelvin (approved):
Approving, thanks for addressing!

laserkelvin:
The merge-queue CI failure (run 35281367167) isn't from this PR's changes — it's a flaky pre-existing test:

`test/data/test_batch.py::TestBatchIsendIrecvTagAlignment::test_tag_sequences_match` → `ValueError: Segment lengths cannot be negative: tensor([-274990144])` from `level_storage.py:689`

**Problem:** the test patches `dist.irecv` with `capture_irecv_tag`, which returns a `MagicMock` without writing into the segment-length buffers. `_BatchRecvHandle.wait()` allocates those buffers with `torch.empty(..., dtype=torch.int32)`, so `SegmentedLevelStorage` validation reads heap garbage. Reproduced locally: 4 failures in 25 runs, each with different garbage (`-1023390736` locally vs `-274990144` in CI).

**Assumed behavior:** the mocked receive should leave the buffers in a state that passes validation, instead of relying on uninitialized memory happening to be non-negative.

**Suggested fix:** one line in `capture_irecv_tag` (line 2931 on current head), matching the pattern `test_custom_send_and_receive_tags_align_in_schema_order` already uses:

```python
def capture_irecv_tag(tensor, *args, **kwargs):
    if "tag" in kwargs:
        recv_tags.append(kwargs["tag"])
    tensor.zero_()
    return MagicMock()
```

Since the file isn't touched by this PR, a tiny follow-up PR to main works too if you'd rather keep this one scoped — but re-queuing without the fix keeps rolling dice. Happy to open it if that's easier.

ys-teh:
/ok to test [da87a8e](https://github.com/NVIDIA/nvalchemi-toolkit/pull/182/commits/da87a8eb16130595545b42658a033eb0c7d6e26c)

laserkelvin:
Last round failed I think because of a corrupted checkpoint download

PR #190, opened 2026-09-10
fix(dynamics): protect LoggingHook snapshots from allocator reuse

### LINKED ISSUE #171: 🐛[BUG]: LoggingHook can log values it never computed (missing record_stream on the D2H copy)
## Version

`0.2.0`, and the same on `main` - `nvalchemi/dynamics/hooks/logging.py` is byte-identical.

## On which installation method(s) does this occur?

`Pip`, `Source`. Not container-specific.

## Describe the issue

If you use `LoggingHook` as a context manager on CUDA, some of the rows it writes contain values
from a different step. They're usually plausible numbers rather than obvious garbage, which is
what makes it nasty - nothing errors, and the CSV looks fine.

I hit it running NPT MD (108-atom FCC Ag, 3 replicas, 900 K, barostat target 0 GPa, MACE
`medium-mpa-0`). The logging is about as ordinary as it gets:

```python
logging_hook = LoggingHook(
    backend="csv",
    log_path=log_path,
    frequency=log_every,                                # 15 in this run
    custom_scalars={"pressure_gpa": pressure_fn},
    stage=DynamicsStage.AFTER_STEP,
)

with logging_hook:
    npt.register_hook(logging_hook)
    batch = npt.run(batch, n_steps=n_prod)
```

The run reported a mean production pressure of **+27.80 GPa**, which can't happen with a
barostat pulling the cell to zero. 7 rows out of 600 looked like this - a normal row first for
comparison:

```csv
step,graph_idx,status,energy,fmax,temperature,pressure_gpa
1500.0,0.0,0.0,-287.3546447753906,2.9172089099884033,915.863037109375,0.7250455617904663
...
1755.0,2.0,0.0,-275.0795593261719,2.6378116607666016,0.0,2121.04443359375
2010.0,0.0,0.0,-282.1431579589844,3.0912575721740723,0.0,2075.7685546875
2010.0,1.0,0.0,-283.6712951660156,3.3936030864715576,0.0,2095.287109375
2010.0,2.0,0.0,-276.0984191894531,2.9171175956726074,0.0,2146.10888671875
2160.0,0.0,0.0,-280.47406005859375,2.591794729232788,926.676025390625,2100.525634765625
```

`pressure_gpa` is holding 2069-2146, which is the *cell volume* in A^3 - the same run's final
per-replica volumes were 2084.7, 2138.9 and 2075.6. `temperature` is zeroed on several of them.
`energy` and `fmax` on those rows are perfectly normal, so nothing about the row flags itself as
suspect. Drop those 7 rows and the mean is **-0.0010 +/- 0.0356 GPa**, which is the right answer.

It wasn't confined to pressure, either. Diffing a full run against a synchronous reference
computed from the same batches at the same stage, 46% of 1500 rows disagreed in at least one
column - `energy` 19.0%, `temperature` 17.8%, `fmax` 11.5%, `status` 4.2%.

## Root cause

`__enter__` creates a side stream, and `_log` then does roughly this:

```python
td = self._compute_columns(batch, step_count, ctx)   # allocated on the MAIN stream
td = _snapshot_tensordict(td)                        # cloned, still MAIN stream

with stream_ctx:                                     # side stream
    self._stream.wait_stream(main_stream)            # orders compute -> copy, correct
    td = td.to("cpu", non_blocking=True)             # queues the D2H, and drops the last
                                                     # reference to the GPU snapshot
self._executor.submit(self._dispatch, td, step_count)
```

The snapshot was allocated against the main stream, so rebinding `td` hands those blocks back
to the main pool. Nothing tells the caching allocator that a side-stream copy is still
outstanding against them, so the next main-stream allocation can take the block and overwrite
it before the copy runs. `wait_stream` orders compute before copy; nothing orders copy before
free. `record_stream` doesn't appear anywhere in the package.

## Proposed fix

Two options. Both are source changes to `_log` (the monkeypatching in the reproducers is only so
one script can show before and after in a single run). I tested both.

### Option A - tell the allocator the side stream is still using the blocks

```python
        with stream_ctx:
            if use_stream:
                self._stream.wait_stream(main_stream)
            gpu_td = td
            td = td.to("cpu", non_blocking=True)
            if use_stream:
                for key in gpu_td.keys():
                    gpu_td[key].record_stream(self._stream)

        self._executor.submit(self._dispatch, td, step_count)
```

Order matters in both directions: `gpu_td = td` has to precede the `.to()` (afterwards the GPU
handle is gone), and `record_stream` has to follow it, since it captures the stream's queue
position and wouldn't otherwise cover the copy. Both mistakes leave code that looks right and
still corrupts, so they're worth a comment if you take this route.

### Option B - keep the snapshot alive until the copy has landed

`_dispatch` already synchronises the side stream before it reads `td`, so holding a reference
across that call is enough. No allocator interaction at all.

```python
        with stream_ctx:
            if use_stream:
                self._stream.wait_stream(main_stream)
            gpu_td = td
            td = td.to("cpu", non_blocking=True)

        self._executor.submit(self._dispatch, td, step_count, gpu_td)
```

```python
    def _dispatch(self, td, step, _gpu_td=None) -> None:
        """``_gpu_td`` is never read. It's the GPU snapshot, held so its blocks can't be
        recycled before the copy of it has completed. Released when this returns."""
        if self._stream is not None:
            self._stream.synchronize()
```

The unused parameter is load-bearing, which is exactly the kind of thing a later cleanup
deletes. Hence the docstring.

### What I tested

Corrupt rows, as shipped vs with each fix, on an A10:

| check | as shipped | + Option A | + Option B |
|---|---|---|---|
| `repro_stream_ownership_deterministic.py`, plain torch | 20 / 20 | 0 / 20 | not tested |
| ...and whether the block was recycled (`data_ptr`) | 20 / 20 | 0 / 20 | not tested |
| `repro_logginghook_deterministic.py` | 75 / 75 | 0 / 75 | 0 / 75 |
| MACE + NPT, `frequency=1`, 1500 rows, no stall | 329 | 0 | not tested |
| MACE + NPT vs a synchronous reference, all columns | 687 (45.8%) | 0 (0.0%) | not tested |

The last row is the one that first convinced me the values were wrong at copy time rather than
compute time: a second hook calling the same `_compute_columns` and reading it back with a
blocking `.cpu()` after `torch.cuda.synchronize()`, diffed row by row on `(step, graph_idx)`.
Rows aligned exactly (0 key mismatches); `pressure_gpa` differed on 24.8%, `energy` 19.0%,
`temperature` 17.8%, `fmax` 11.5%, `status` 4.2%, `volume_a3` 2.8%. Sub-0.1% reduction-order
jitter was counted separately and excluded.

### What it costs

3000 `_log` calls, 3 graphs, 8 columns, `frequency=1`, 5 interleaved repeats:

| variant | median us/call | min | max |
|---|---|---|---|
| as shipped | 1359.1 | 1336.9 | 1376.5 |
| Option A (`record_stream`) | 1399.8 | 1389.0 | 1424.7 |
| Option B (hold the reference) | 1369.3 | 1337.6 | 1373.9 |

Baseline's own run-to-run spread was 2.91%, so the percentages are near the noise - but Option
A's range never overlaps the baseline's, so its ~40 us/call is real, while Option B's sits
inside it. Peak memory was 0.02 MiB for all three. My baseline includes synthetic batch
construction inside the timed loop, so the *percentages* flatter the fix; the absolute +40
us/call is the figure that carries over, which against a 10-50 ms MD step is +0.1-0.4% at
`frequency=1`. Option A's cost is per tensor so it grows with column count; Option B's doesn't.

I'd lean towards A. It keeps `_log` correct on its own terms, whereas B's correctness depends
on `_dispatch` continuing to synchronise first - if that's ever refactored into an event-based
wait, B stops protecting anything and no test fails. But 40 us is 40 us and it's your hot path.

## How likely you are to hit it

The odds depend on allocator pressure and logging frequency, but the ownership error is there
either way. Same script on an A10, one variable at a time:

| config | corrupt rows |
|---|---|
| `frequency=15`, one custom scalar (my original MD run) | 7 / 600 |
| `frequency=1`, one custom scalar | 32 / 1500 |
| `frequency=1`, two scalars, both trivial | 44 / 3000 |
| `frequency=1`, two scalars, one allocates scratch | 602 / 3000 |
| MACE + NPT, `frequency=1` | 647 / 3000 |
| MACE + NVE, NVT-Langevin, or NVT-Nose-Hoover | 0 / 3000 (each) |
| Lennard-Jones + NPT | 0 / 1500 |

The D2H is only ~1 KB and usually lands before anything can recycle the block. What opens the
window is allocator pressure: a custom scalar that allocates scratch of its own takes the rate
from ~1.5% to ~20%. Those two rows are matched (both log two scalars and check both columns,
only the second scalar's body differs).

Logging less often reduces exposure but does not remove it: the run that first showed me this
logged every 15 steps and still lost 7 rows in 600. The 0 / 1500 rows reproduce, so they aren't sampling noise - but they mean nothing happens to land
in the window in those configurations, not that `_log` is correct there.

The scratch-allocating case isn't contrived, either. `compute_pressure_tensor` takes `kinetic_scratch`
and `pressure_scratch` as required arguments, so a custom scalar that logs pressure using the
toolkit's own kernels allocates and frees several buffers on every call by construction. The
configuration most exposed to this is the one the documented API leads you to.

One thing I cannot explain: among integrators, only NPT shows it. Same model, same logging,
same scalars. I ruled out the obvious candidates - NVE with `stress` forced into
`active_outputs` is clean, so it isn't the model's stress path, and NVT-Nose-Hoover with
`chain_length=3` is clean, so it isn't per-graph thermostat state either. What is left that is
unique to NPT is the barostat mutating `batch.cell` each step, the `nvalchemiops` NPT kernels,
and `NeighborListHook(skin=0.0)` therefore doing real work every step instead of rebuilding an
unchanged list. I have no evidence to choose between those and would rather report the
observation than guess. It does not affect the diagnosis: the deterministic reproducers below
use no integrator at all.

## Minimum reproducible example

Two scripts, neither needs a model or checkpoint. Both reproduce on every run I've done.

Both use `torch.cuda._sleep` to hold the side stream, so the copy is queued behind a long spin
and hasn't run when the block is recycled. Stream ordering guarantees the copy sits behind the
spin; what isn't guaranteed is that the host reaches the racing allocation inside that window,
so this is a wide margin rather than a proof. The stall is calibrated to wall-clock
milliseconds the way `get_cycles_per_ms` does in your own test suite, so it doesn't shrink on a
faster GPU. Measured on an A10: a 500 ms window against a host that reaches the racing
allocation in ~95 us median, 24 ms worst case. Both scripts assert, so they fail loudly rather
than passing quietly if that margin is ever too thin - raise `STALL_MS` if so.

If you'd want a hard guarantee rather than a margin, `cuStreamWaitValue32` would give one (the
stream blocks until the host writes a flag), but it isn't exposed through PyTorch's Python API,
so it would mean ctypes against libcuda or a `load_inline` extension. Happy to provide that
version if you'd prefer it.

The stall is what makes them repeatable, not what causes the bug. The MACE+NPT numbers above
involve no stall at all.

Both print 100% corrupt on `main` and 0% with `record_stream` applied.

### 1. `repro_stream_ownership_deterministic.py`

The mechanism in plain torch, no toolkit import. The `data_ptr` check lets you watch the
allocator hand the freed block straight to the next allocation.

```python
#!/usr/bin/env python
"""The stream-ownership bug behind LoggingHook, reduced to plain torch.

A tensor is allocated on the main stream, a copy to CPU is queued on a side stream, then the
last reference is dropped. The allocator hands that block to the next main-stream allocation
while the copy is still pending, so the copy ends up reading whatever overwrote it.

`torch.cuda._sleep` holds the side stream busy so the copy definitely hasn't run when the
block is recycled. That's what makes it deterministic; it isn't what causes the bug.

The data_ptr check shows the allocator's decision directly: without record_stream the poison
tensor lands on the freed block every time, with it, never.

    python repro_stream_ownership_deterministic.py
"""
import sys

import torch

if not torch.cuda.is_available():
    sys.exit("SKIP: needs a CUDA device")

DEV = torch.device("cuda")
N = 256              # small: lands in the caching allocator's small-block pool
ITERS = 20
STALL_MS = 500.0     # calibrated below; the D2H it must outlast is ~1 KB
GOOD, POISON = 1.0, -999.0

def _cycles_for_ms(ms):
    """Cycles for torch.cuda._sleep to stall ~ms milliseconds ON THIS GPU.

    Same calibration PyTorch's own test suite uses (get_cycles_per_ms in
    torch.testing._internal.common_utils), inlined here so this depends only on public API.
    A raw cycle count is not portable: 1e9 cycles is ~590 ms on an A10 and much less on a
    faster part, which would silently shrink the margin this relies on.
    """
    import time
    torch.cuda.synchronize()
    t0 = time.perf_counter()
    torch.cuda._sleep(1_000_000)
    torch.cuda.synchronize()
    return int(ms * 1_000_000 / ((time.perf_counter() - t0) * 1000))


print(f"torch  {torch.__version__}")
print(f"device {torch.cuda.get_device_name(0)}")


def one_round(use_record_stream: bool) -> tuple[int, int, int]:
    main = torch.cuda.current_stream(DEV)
    side = torch.cuda.Stream()
    corrupted = same_block = 0
    for _ in range(ITERS):
        src = torch.full((N,), GOOD, device=DEV, dtype=torch.float32)
        ptr = src.data_ptr()

        with torch.cuda.stream(side):
            side.wait_stream(main)          # compute -> copy ordering (already correct upstream)
            torch.cuda._sleep(_cycles_for_ms(STALL_MS))         # hold the side stream: the copy cannot run yet
            dst = src.to("cpu", non_blocking=True)
            if use_record_stream:
                src.record_stream(side)     # the fix: block is in use by `side`

        del src                             # refcount drops; block returns to the MAIN pool

        # The next same-size main-stream allocation. Without record_stream the allocator
        # believes this block is free and hands it straight back.
        poison = torch.full((N,), POISON, device=DEV, dtype=torch.float32)
        if poison.data_ptr() == ptr:
            same_block += 1

        side.synchronize()                  # now the copy has certainly run
        if float(dst[0]) != GOOD:
            corrupted += 1
        del poison, dst
    return corrupted, same_block, ITERS


for label, fix in (("as shipped (no record_stream)", False), ("with record_stream", True)):
    bad, reused, total = one_round(fix)
    print(f"\n{label}")
    print(f"  poison landed on the freed block : {reused}/{total}")
    print(f"  copied-out value was corrupted   : {bad}/{total}")
    if fix:
        print("  (0/20 reuse is the fix working: the allocator withholds the block)")
    elif reused == 0:
        print("  SETUP DID NOT HOLD - allocator never recycled the block; not evidence either way")

print()
bad_before, _, total = one_round(False)
bad_after, _, _ = one_round(True)
print(f"VERDICT before={bad_before}/{total} after={bad_after}/{total}")
assert bad_after == 0, "record_stream did not fix it"
assert bad_before == total, (
    f"expected deterministic corruption, got {bad_before}/{total} - the stall may be too "
    f"short for this GPU; raise STALL_MS"
)
print("result: every iteration corrupted without the fix, none with it")
```

### 2. `repro_logginghook_deterministic.py`

The real `LoggingHook`, nothing patched in the failing run. The logged `marker` column is a
custom scalar defined to return the step number, so every row's correct value is known without
reading the GPU - the check never synchronises, which matters because a stray synchronise hides
this bug.

```python
#!/usr/bin/env python
"""Deterministic reproduction against the real, unmodified LoggingHook. No model needed.

`custom_scalars` callables run inside `_compute_columns`, which happens before `_log` enters
its `with torch.cuda.stream(...)` block. So a scalar can queue a stall on the hook's own side
stream from outside the library, putting it ahead of the D2H copy. Nothing is patched in the
failing run; `_patch()` only exists to show the fix in the second run.

    pip install "nvalchemi-toolkit[cu12] @ git+https://github.com/NVIDIA/nvalchemi-toolkit.git@main"
    python repro_logginghook_deterministic.py

Expected: every row has marker == step. Actual: every row is wrong.

FIX=record_stream (default) or FIX=hold_ref picks which candidate fix the second run applies.
Both take it to 0/75.
"""
import csv
import os
import sys
import tempfile
from pathlib import Path

import torch

if not torch.cuda.is_available():
    sys.exit("SKIP: needs a CUDA device")

import nvalchemi  # noqa: E402
from nvalchemi.dynamics.hooks import LoggingHook  # noqa: E402

DEV = torch.device("cuda")
N_GRAPHS, N_ATOMS, N_STEPS = 3, 32, 25
                            # Measured margin: the host reaches the racing allocation in
                            # ~95 us median, 24 ms worst case, so this leaves ~25x even in
                            # the worst case seen. Raise it if the assertion below trips.

print(f"torch     {torch.__version__}")
print(f"nvalchemi {getattr(nvalchemi, '__version__', '?')}")
print(f"device    {torch.cuda.get_device_name(0)}")


class FakeBatch:
    """Only the attributes `_compute_columns` reads."""

    def __init__(self, step):
        n = N_ATOMS * N_GRAPHS
        self.device, self.num_graphs = DEV, N_GRAPHS
        self.status = torch.zeros(N_GRAPHS, device=DEV)
        self.energy = torch.full((N_GRAPHS, 1), -100.0 - step, device=DEV)
        self.forces = torch.full((n, 3), 0.1, device=DEV)
        self.batch_idx = torch.arange(n, device=DEV) // N_ATOMS
        self.velocities = torch.full((n, 3), 0.01, device=DEV)
        self.atomic_masses = torch.full((n,), 12.0, device=DEV)
        self.num_nodes_per_graph = torch.full((N_GRAPHS,), N_ATOMS, device=DEV,
                                              dtype=torch.long)


class Ctx:
    def __init__(self, batch, step):
        self.batch, self.step_count = batch, step



def _cycles_for_ms(ms: float) -> int:
    """Cycles for `torch.cuda._sleep` to stall roughly `ms` milliseconds ON THIS GPU.

    Same calibration PyTorch's own test suite uses (`get_cycles_per_ms` in
    torch.testing._internal.common_utils), inlined so this script depends only on public API.
    A raw cycle count is not portable: 1e9 cycles is ~590 ms on an A10 and much less on a
    faster part, which would silently shrink the margin the reproducer relies on.
    """
    import time
    torch.cuda.synchronize()
    t0 = time.perf_counter()
    torch.cuda._sleep(1_000_000)
    torch.cuda.synchronize()
    per_ms = 1_000_000 / ((time.perf_counter() - t0) * 1000)
    return int(ms * per_ms)


STALL_MS = 500.0

def run(use_record_stream: bool) -> tuple[int, int]:
    out = Path(tempfile.mkdtemp()) / "log.csv"
    hook_box = {}

    def marker(ctx):
        """Correct value is the step number. Also stalls the hook's own side stream."""
        hook = hook_box.get("hook")
        if hook is not None and hook._stream is not None:
            with torch.cuda.stream(hook._stream):
                torch.cuda._sleep(_cycles_for_ms(STALL_MS))
        return torch.full((N_GRAPHS,), float(ctx.step_count), device=DEV)

    with LoggingHook(backend="csv", frequency=1, log_path=str(out),
                     custom_scalars={"marker": marker}) as hook:
        hook_box["hook"] = hook
        if use_record_stream:
            _patch(hook, FIX)
        for step in range(N_STEPS):
            batch = FakeBatch(step)
            hook(Ctx(batch, step), None)
            # Stand-in for the next MD step: small main-stream allocations, the size class the
            # snapshot columns live in. One of these takes the block the copy still needs.
            junk = [torch.full((N_GRAPHS,), -999.0, device=DEV) for _ in range(64)]
            del junk, batch

    rows = list(csv.DictReader(open(out)))
    bad = sum(1 for r in rows if float(r["marker"]) != float(r["step"]))
    return bad, len(rows)


def _patch(hook, kind="record_stream"):
    """A candidate fix, applied to this instance only."""
    import contextlib

    from nvalchemi.dynamics.hooks.logging import _snapshot_tensordict

    def _log(batch, step_count, ctx, _self=hook):
        device = batch.device
        use_stream = _self._stream is not None and device.type == "cuda"
        td = _snapshot_tensordict(_self._compute_columns(batch, step_count, ctx))
        ctxm = (torch.cuda.stream(_self._stream) if use_stream
                else contextlib.nullcontext())
        with ctxm:
            if use_stream:
                _self._stream.wait_stream(torch.cuda.current_stream(device))
            gpu_td = td
            td = td.to("cpu", non_blocking=True)
            if use_stream:
                for key in gpu_td.keys():
                    gpu_td[key].record_stream(_self._stream)
        _self._executor.submit(_self._dispatch, td, step_count)

    def _log_hold(batch, step_count, ctx, _self=hook):
        device = batch.device
        use_stream = _self._stream is not None and device.type == "cuda"
        td = _snapshot_tensordict(_self._compute_columns(batch, step_count, ctx))
        ctxm = (torch.cuda.stream(_self._stream) if use_stream
                else contextlib.nullcontext())
        with ctxm:
            if use_stream:
                _self._stream.wait_stream(torch.cuda.current_stream(device))
            gpu_td = td
            td = td.to("cpu", non_blocking=True)
        # `_dispatch` synchronises the side stream first, so holding gpu_td across the call
        # keeps the blocks reserved until the copy has landed.
        _self._executor.submit(
            lambda a, b, _keep: _self._dispatch(a, b), td, step_count, gpu_td)

    hook._log = _log if kind == "record_stream" else _log_hold


FIX = os.environ.get("FIX", "record_stream")
bad_before, total = run(False)
print(f"\nas shipped          {bad_before}/{total} rows wrong "
      f"({100 * bad_before / max(total, 1):.0f}%)")
bad_after, total_after = run(True)
print(f"with {FIX:18s}{bad_after}/{total_after} rows wrong "
      f"({100 * bad_after / max(total_after, 1):.0f}%)")

assert bad_after == 0, "record_stream did not fix it"
assert bad_before == total, (
    f"expected every row wrong, got {bad_before}/{total}; raise STALL_MS if this GPU is fast"
)
print("\nresult: every logged row corrupted without the fix, none with it")
```

There's a third script behind the MACE + NPT numbers quoted above - the same check run
against a real workload with no artificial stall. I left it out because it downloads a
checkpoint and takes a few minutes, which felt out of scope for a minimal reproducer, but I'm
happy to attach it if it's useful.

## Relevant log output

`repro_stream_ownership_deterministic.py` - the `data_ptr` line is the allocator's own decision:
without the fix it hands the freed block straight back, with it the block is withheld.

```shell
$ python repro_stream_ownership_deterministic.py
torch  2.13.0+cu130
device NVIDIA A10

as shipped (no record_stream)
  poison landed on the freed block : 20/20
  copied-out value was corrupted   : 20/20

with record_stream
  poison landed on the freed block : 0/20
  copied-out value was corrupted   : 0/20
  (0/20 reuse is the fix working: the allocator withholds the block)

VERDICT before=20/20 after=0/20
result: every iteration corrupted without the fix, none with it
```

`repro_logginghook_deterministic.py`, against the real hook:

```shell
$ FIX=record_stream python repro_logginghook_deterministic.py
torch     2.13.0+cu130
nvalchemi 0.2.0
device    NVIDIA A10
as shipped          75/75 rows wrong (100%)
with record_stream     0/75 rows wrong (0%)
result: every logged row corrupted without the fix, none with it

$ FIX=hold_ref python repro_logginghook_deterministic.py
torch     2.13.0+cu130
nvalchemi 0.2.0
device    NVIDIA A10
as shipped          75/75 rows wrong (100%)
with hold_ref          0/75 rows wrong (0%)
result: every logged row corrupted without the fix, none with it
```

And the same defect in a real run with no artificial stall - MACE `medium-mpa-0` + NPT at
`frequency=1`, custom scalar defined to return the step number:

```shell
$ python repro_mace_npt.py
torch      2.13.0+cu130
nvalchemi  0.2.0
device     NVIDIA A10
logged rows        1500
rows != their step 329  (21.9%)
   step=1 graph_idx=0 marker=1.4013e-45 scratch=1  (both should == step)
   step=1 graph_idx=1 marker=2.8026e-45 scratch=1  (both should == step)
   step=1 graph_idx=2 marker=4.2039e-45 scratch=1  (both should == step)
REPRODUCED
AssertionError: 329/1500 logged rows carry scalars that do not match the step
exit=1

$ NVALCHEMI_RECORD_STREAM=1 python repro_mace_npt.py
torch      2.13.0+cu130
nvalchemi  0.2.0
device     NVIDIA A10
record_stream patch: APPLIED
logged rows        1500
rows != their step 0  (0.0%)
clean
exit=0
```

Those three marker values are the integers 1, 2 and 3 read as float32 - `graph_idx` (int64)
seen through a float column.

## Environment details

```shell
+ Environment location: Bare-metal (Slurm compute node)
+ GPU: NVIDIA A10
+ NVIDIA driver: 595.58.03
+ CUDA version (torch): 13.0
+ Python version: 3.12.3
+ PyTorch version: 2.13.0+cu130
+ NVIDIA Warp version: 1.16.0
+ nvalchemi-toolkit: 0.2.0, pip install from git+https://github.com/NVIDIA/nvalchemi-toolkit.git@main
```

FabioFalcioni:

## Description

`LoggingHook` performs non-blocking CUDA-to-host copies on a dedicated CUDA stream. The source snapshot tensors could be released immediately after a copy was queued, allowing the CUDA caching allocator to reuse their storage before the copy completed. As reported in #171, this can silently mix values from another step into logged rows, producing plausible but incorrect results.

Record each CUDA snapshot tensor on the logging stream after enqueueing its copy. This keeps the storage unavailable for reuse until the transfer completes while preserving asynchronous logging behavior.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Related Issues

Fixes #171

## Changes Made

- Record CUDA snapshot tensors on the logging stream after enqueueing the non-blocking CPU transfer.
- Skip non-CUDA snapshot tensors when recording the CUDA stream.
- Add a CUDA regression test that verifies snapshot tensors are recorded on the logging stream.

## Testing

- Observer-hook module: 78 passed on NVIDIA A10; 44 passed and 34 CUDA-dependent tests skipped locally.
- The targeted CUDA regression failed against the unmodified implementation with zero `record_stream` calls instead of six, then passed with the fix.
- Changed-file pre-commit checks passed.
- [ ] Linting passes (`make lint`)

## Checklist

## Additional Notes

The public API and documented asynchronous behavior are unchanged. A10 validation used PyTorch 2.13.0+cu130.

> [!TIP]
> of the code, nor is it an indication that the PR will be accepted/rejected.


laserkelvin (approved):
LGTM

laserkelvin:
/ok to test [12202b9](https://github.com/NVIDIA/nvalchemi-toolkit/pull/190/commits/12202b9fef26b11af2eb5d1e9cf1741714157973)

laserkelvin:
Forcing merge, same multiGPU CI test issue

PR #193, opened 2026-09-15
fix: device- and dtype-robust segment storage, fp32 loss accumulation, and resumable device moves

EricZQu:

## Description

Six independent core defects, each a confirmed crash or wrong numerical result on `main`
with a GPU or CPU reproduction. They cluster into three themes: segmented level storage was
neither device- nor dtype-robust (the Warp expansion kernel launched against the current CUDA
device rather than the pointer's, the storages recorded the requested device instead of the
resolved one, and the bulk merge left `segment_lengths` behind); graph-balanced losses
accumulated scatter sums in the residual dtype, so low-precision reductions saturated; and
device moves did not survive a resume, with live checkpoint restores loading onto the recorded
device and `run()` reusing optimizer state left on the pre-move device. Every change is one
commit with its own regression tests, and no behavior changes beyond the fixes. Batch indexing
now works with any current CUDA device.

Rebased onto #194. Two of the storage fixes this branch originally carried — the int32 pointer
cast in the Warp expansion kernel and the `segment_lengths` move in
`SegmentedLevelStorage.concatenate` — landed there independently, so only their regression
coverage remains here.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update

## Related Issues

None filed; all were found while integrating against `main`.

## Changes Made

- **`test(data)`: cover int32 batch pointers in Warp segment expansion** — regression
  coverage for `Cannot convert Torch type torch.int32 to Warp type int64` out of
  `Batch.index_select` on any batch cloned or moved to CUDA after its pointer was
  materialized. The cast itself landed in #194.
- **`fix(data)`: launch segment expansion on the pointer's device** — removes
  `CUDA error: an illegal memory access was encountered` whenever the current device is not
  `cuda:0`, and stops the launch from silently leaving its own device current for the rest of
  the process.
- **`fix(data)`: record the resolved device on level storages** — removes
  `Expected all tensors to be on the same device` from lazily built batch pointers and
  `torch.cat` after a storage was moved with a bare `.to("cuda")`.
- **`test(data)`: cover appending a CPU storage into an accelerator one** — regression
  coverage for the `wrapper_CUDA_cat` device error out of `Batch.append` when a CPU batch is
  folded into one already on the accelerator. The move itself landed in #194.
- **`fix(training)`: accumulate per-graph scatter sums in fp32** — removes the silent
  truncation that made a per-atom-normalized bf16 force loss over 3000 atoms 3.6x too large
  (10% in fp16); fp32/fp64 results are bit-identical.
- **`fix(models)`: write demo node embeddings to the atoms group** — removes the
  `Length mismatch` that made `DemoModelWrapper.compute_embeddings` raise on every `Batch`,
  one-graph batches included.
- **`fix(training)`: move models to strategy devices in `validate()`** — removes the
  CPU-model/CUDA-batch device error from a standalone validation pass taken before `run()`.
- **`fix(training)`: load live restores onto the live device and re-home optimizer state** —
  removes `Tensors of the same index must be on the same device` and
  `Expected all tensors to be on the same device, but found at least two devices` from
  resuming a checkpoint onto a different device (including a `DDPHook` re-pinning rank > 0).
  Adds the public `nvalchemi.training.rehome_optimizer_state` helper, applied automatically
  by `run()`.

Review follow-ups, each a separate commit on top:

- **`fix(training)`: keep half-precision per-graph sums in fp32 through normalization** —
  half-precision graph-balanced losses no longer overflow after the fix above widened
  accumulation: the widened total was cast straight back to fp16, and a per-graph sum of
  squared force errors over a few thousand atoms passes fp16's 65504 ceiling, so the loss
  returned `inf` even though the normalized mean it was about to compute is finite.
  `per_graph_sum` / `per_graph_mean` now return fp32 for fp16 and bf16 inputs (fp32 and fp64
  are bit-identical as before), and the graph-balanced force losses divide fp32 by fp32.
- **`fix(training)`: re-home resumed optimizer state on every reuse, including `train_batch()`** —
  driving a resumed strategy one batch at a time across a device change now works. Rehoming
  was wired into `run()` only, so `train_batch()` still handed back optimizer state stranded
  on the pre-move device; it now lives in the optimizer-reuse path every entry point shares.
- **`fix(training)`: re-home nested optimizer state containers** — a custom optimizer that
  keeps its moments inside a dict, list, or tuple is rehomed as thoroughly as `Adam`, instead
  of having only its top-level tensors follow the parameters. Container types are preserved
  and non-tensor values are left alone.
- **`fix(data)`: keep `Batch.device` in step with its storage's resolved device** — the
  resolution above stopped at the storage, so a batch built or moved with a bare `cuda` still
  recorded `cuda` while its storage recorded `cuda:N`. Batch-level allocations follow
  `Batch.device`, so with the storage on `cuda:1` and device 0 current `batch[idx]` raised
  `indices should be either on cpu or on the same device as the indexed tensor (cuda:1)` and
  `edge_ptr` came back on `cuda:0`. `Batch` now records through the same `_resolve_device`
  helper the storages use, and `Batch.to` adopts the storage's device after the move.
- **`fix(training)`: keep padded force-loss sums in fp32 for half-precision inputs** — the
  fp16-ceiling fix above reached the dense `(V, 3)` layout only, because the padded
  `(B, V_max, 3)` branches sum without `per_graph_sum`. On the same 3000-atom fp16 input the
  dense layout returned 16.1244 and the padded one `inf` (`ForceL2NormLoss`: 64.457 against
  `inf`), and half-precision inputs returned fp32 on one layout and fp16 on the other. Both
  sides of every force reduction now sum in at least fp32, so the two layouts agree; fp32 and
  fp64 stay bit-identical.
- **`fix(models)`: pool every embedding feature in the demo graph embedding** — the graph
  embeddings `DemoModelWrapper.compute_embeddings` returns came back with every feature but
  the first left at zero. `scatter_add_` does not broadcast the `(N, 1)` batch index over the
  `(N, H)` node embeddings, so only feature 0 was pooled; the index is now expanded to the
  source's shape.
- **`fix(models)`: register demo node embeddings through the public node-level path** — the
  demo wrapper wrote node embeddings straight into the atoms group, which places the tensor
  but leaves the key unregistered, so the next public `batch.node_embeddings = ...` fell back
  to the system group and left the same name at two levels. It now inserts through
  `Batch.add_key`, the public path that registers the field with the storage's attribute map.
- **`fix(training)`: widen dense force-component sums before graph reduction** — the dense
  graph-balanced force losses still collapsed an atom's three Cartesian components in the
  input dtype before the widening reduction, so a single fp16 residual near 150 returned
  `inf` where the padded layout returned a finite value. Both dense sums now accumulate in
  fp32, and `ForceHuberLoss` inherits the fix.
- **`fix(training)`: place the inference model before standalone validation** — `validate()`
  moved the training models onto `devices` but not the published inference/EMA slot it runs
  instead when `use_ema` selects it, so a slot filled before `devices` changed met batches on
  a device it had never been moved to. The slot is now placed alongside the models, the same
  way `set_inference_model` places it.
- **`fix(data)`: derive `Batch.device` from a supplied storage** — `Batch(storage=...,
  device="cuda")` resolved the bare request against the current GPU, so a storage held on
  another GPU left `Batch.device` disagreeing with its own data and batch-level allocations
  built for the wrong device. A supplied storage now decides the recorded device; an indexed
  device it is not on raises `ValueError`.
- **`fix(data)`: move `segment_lengths` to the target device in bulk merges** — the move
  #194 made in `SegmentedLevelStorage.concatenate` reached the single-storage path only.
  `MultiLevelStorage.from_batches`, and `MultiLevelStorage.concatenate` which delegates to it,
  still concatenated each group's segment lengths where they already were, so merging storages
  held on different devices raised out of `torch.cat`. They are now moved like every other
  tensor.
- **`fix(training)`: report the live strategy's devices in restored metadata** — restoring
  into a live strategy returned `strategy_metadata` built from the raw `map_location`, which
  could name a device none of the restored models were on. On that path `map_location` only
  stages the load — the live strategy's `devices` decide where everything comes to rest — so
  the metadata now reports those devices, and `load_checkpoint` documents the rule.
- **`fix(data)`: route an attribute write to the group that already holds it** — a tensor
  assigned to a `Batch` resolved its level from the attribute map alone and fell back to the
  system group for anything the map did not know, so a per-node tensor a model wrapper had
  placed in the atoms group was shadowed by a system-level copy on the next
  `batch.<key> = tensor`, and `to_data_list()` then raised. A write now follows the group that
  already holds the key.
- **`fix(data)`: allocate a `Batch`'s own storage on the requested device** — `Batch(device=
  "cuda:1")` with no storage built its storage on CPU, so the batch reported `cuda:1` while
  every tensor assigned to it landed on the host. The storage is now built on the requested
  device.
- **`fix(data)`: keep buffer writes on the destination device** — `Batch.put` with a source on
  another device, and therefore `GPUBuffer.write` with a CPU batch, took down the process with
  a segmentation fault: the fit-mask kernels launched on the source's device against
  destination pointers. `GPUBuffer.write` now moves an incoming batch to the buffer's device
  as `HostMemory.write` already moves its items to CPU, `Batch.put` raises on a source held
  elsewhere, and the put kernels launch on the destination and refuse a mixed-device pair.
- **`fix(models)`: register MACE node embeddings through the public node-level path** —
  `MACEWrapper.compute_embeddings` carried the same private-group write the maintainer flagged
  in the demo wrapper, with the same consequence for a later reassignment. It now uses the
  public registering path.
- **`fix(training)`: accumulate validation losses in float64 at update time** — the validation
  loss accumulator kept its running sums in the loss's dtype and widened only when building
  the summary, so a bf16 running sum stopped growing part-way through an epoch and reported a
  validation loss 36% below the truth. Every running sum is widened as it is taken.
- **`fix(training)`: widen the default loss reduction for half-precision residuals** — the
  validity-weighted mean every loss leaf inherits summed in the residual's dtype, so a finite
  fp16 energy residual over enough graphs returned `inf`. It now accumulates in at least fp32,
  like the force terms; fp32 and fp64 stay bit-identical.

Deliberately out of scope and left for a separate discussion: keying
`models/pipeline.py`'s captured neighbor sources by owning pipeline.

## Testing

## Checklist

## Additional Notes

- The commits are independently revertable and ordered so the two kernel-boundary changes
  land first; the device fix removes the now-unused `device` argument from the private
  `_expand_segments_warp` helper rather than leaving it silently ignored.
- `_resolve_device` changes one observable edge: constructing a level storage with
  `device="cuda"` on a CUDA-less host now fails at construction rather than at the first
  tensor move.
- `rehome_optimizer_state` is deliberately public and exported from `nvalchemi.training`;
  it pairs with the existing `move_to_devices` for anyone moving models by hand.
- `TrainingStrategy.validate()` now honours `devices`, so a caller who relied on models
  staying on CPU while `devices` said CUDA will see them moved. That is the documented
  contract of `devices`, and `run()` already did it.
- Two contract changes worth a second look. `Batch.put` now raises on a source held on another
  device instead of segfaulting; `GPUBuffer.write` moves rather than raising, because that is
  what `HostMemory.write` already does for its own device. And `Batch(storage=...)` rejects an
  indexed device the storage is not on, rather than recording a disagreement.

> [!TIP]

🤖 Generated with [Claude Code](https://claude.com/claude-code)


EricZQu:
@laserkelvin Here are all the bugs I found. It seems I cannot assign reviewers so just tagging you here. The distillation PR is currently using walkarounds for these.

laserkelvin:
/ok to test [df694e9](https://github.com/NVIDIA/nvalchemi-toolkit/pull/193/commits/df694e935a00420c5a73362a43b2cc0f641c767c)

laserkelvin (changes requested):
Found five blocking correctness bugs and two smaller device/bookkeeping problems. Inline comments include minimum reproducers.

nvalchemi/models/demo.py
```diff
@@ -246,7 +246,13 @@ def compute_embeddings(
         graph_embedding.scatter_add_(0, batch_indices.unsqueeze(-1), embedding)
         # write embeddings to data structure
         data.graph_embeddings = graph_embedding
-        data.node_embeddings = embedding
+        # A plain attribute set on a Batch routes to the system group, where a
+        # per-node tensor fails the batch-size check.
+        atoms_group = data._atoms_group if isinstance(data, Batch) else None
```
laserkelvin: blocking: This method returns graph embeddings computed with an `(N, 1)` scatter index and an `(N, H)` source. `scatter_add_` does not broadcast that index, so only feature 0 is pooled; the node embeddings stored here are correct but `graph_embeddings` are not.

Minimum reproducer:
```python
out = torch.zeros(2, 4)
out.scatter_add_(0, torch.tensor([[0], [0], [1]]), torch.arange(12.).reshape(3, 4))
# columns 1..3 remain zero
```

Expected `(B, H)` embedding sums every feature per graph. Expand earlier index with `batch_indices[:, None].expand_as(embedding)`, then add manual graph-sum value test.

nvalchemi/models/demo.py
```diff
@@ -246,7 +246,13 @@ def compute_embeddings(
...
+        if atoms_group is not None:
```
laserkelvin: issue: Writing through `_atoms_group` bypasses `Batch.keys` and storage attribute map. Later public `batch.node_embeddings` assignment routes unknown field to system group.

Minimum reproducer:
```python
batch = model.compute_embeddings(batch)
batch.node_embeddings = torch.zeros(batch.num_nodes, hidden_dim)
# ValueError: Length mismatch: num_nodes vs num_graphs
```

Use public node-level insertion path that registers field, then add reassignment regression.

nvalchemi/training/losses/terms.py
```diff
@@ -593,6 +593,15 @@ class ForceMSELoss(BaseLossFunction):
         How to handle prediction/target dtype mismatches before validation.
         ``"strict"`` raises; the other policies cast one tensor to match the
         other.
+
```
laserkelvin, line 596: blocking: New guarantee is not true for dense, graph-balanced `ForceMSELoss`. That path sums xyz in fp16 before `per_graph_sum` widens it; padded path widens first.

Minimum reproducer: one dense fp16 atom with residual `(150, 150, 150)` has finite component squares, but `squared_error.sum(dim=-1)` is `inf`; dense loss becomes `inf` while padded loss is finite.

Sum dense xyz numerators and valid counts with `dtype=acc_dtype` before `per_graph_sum`, and test this one-atom dense/padded pair.

nvalchemi/training/strategy.py
```diff
@@ -2063,6 +2080,7 @@ def validate(self) -> dict[str, Any] | None:
             raise RuntimeError(
                 "TrainingStrategy.validate() requires a validation_config."
             )
+        self.models = move_to_devices(self.models, self.devices)
```
laserkelvin, line 2087: blocking: `validate()` moves `self.models`, but validation can run `inference_model` when `use_ema="auto"`. Published EMA model can remain on old device.

Minimum reproducer:
```python
strategy.set_inference_model(ema_cpu)
strategy.devices = [torch.device("cuda")]
strategy.validate()  # CUDA batch reaches CPU EMA: device mismatch
```

Move selected inference model to `self.devices[0]` before validation loop, then add this device-change case.

nvalchemi/data/batch.py
```diff
@@ -238,11 +242,7 @@ def __init__(
             self, "_storage", storage if storage is not None else MultiLevelStorage()
         )
         object.__setattr__(self, "_data_class", AtomicData)
-        object.__setattr__(
-            self,
-            "device",
-            torch.device(device) if isinstance(device, str) else device,
-        )
+        object.__setattr__(self, "device", _resolve_device(device))
```
laserkelvin: blocking: Resolving bare `"cuda"` independently still disagrees with supplied storage on another GPU. `Batch.device` becomes current `cuda:0` even when `storage` is on `cuda:1`; `index_select()` then mixes devices.

Minimum reproducer on two GPUs:
```python
torch.cuda.set_device(0)
batch = Batch(storage=storage_on_cuda_1, device="cuda")
batch.index_select(torch.tensor([0], device="cuda:0"))  # device mismatch
```

When storage is supplied, derive `Batch.device` from `storage.device` or reject conflict. Apply same rule in `_construct`.

nvalchemi/data/level_storage.py
```diff
@@ -1727,7 +1770,9 @@ def concatenate(
             batch_size=[new_total],
             device=self.device,
         )
-        self.segment_lengths = torch.cat([self.segment_lengths, other.segment_lengths])
+        self.segment_lengths = torch.cat(
```
laserkelvin: blocking: This fixes `SegmentedLevelStorage.concatenate`, but `MultiLevelStorage.from_batches()` still concatenates raw `g.segment_lengths` before moving them to destination device. `MultiLevelStorage.concatenate()` delegates there.

Minimum reproducer: CPU segmented storage plus CUDA segmented storage passed to `MultiLevelStorage.from_batches(...)` raises from mixed-device `torch.cat`.

Move every `g.segment_lengths` to `device` before bulk `torch.cat`, and cover both bulk entry points.

nvalchemi/training/_checkpoint.py
```diff
@@ -1862,13 +1862,16 @@ def load_checkpoint(
...
             )
+        # A live restore targets the live strategy, not the recorded device.
         loaded = _restore_checkpoint_into_strategy(
             root,
             manifest,
             checkpoint_index=checkpoint_index,
             strategy=strategy,
             strategy_metadata=strategy_metadata,
-            map_location=load_location,
+            map_location=(
```
laserkelvin, line 1890: issue: Live restore targets strategy device, but returned metadata is later overwritten from `map_location`. Metadata then disagrees with restored models.

Minimum reproducer: restore into CPU strategy with `map_location="cuda:0"`; model and `strategy.devices` stay CPU, while returned metadata reports `cuda:0`.

For live-strategy path, set returned metadata from `strategy.devices`; keep `map_location` override only for reconstructed strategies.

laserkelvin (commented):
Inline reproducers for requested changes.

EricZQu:
Thanks for the review! @laserkelvin Every point is fixed and found a few more, updated in the PR body

laserkelvin:
/ok to test [d196242](https://github.com/NVIDIA/nvalchemi-toolkit/pull/193/commits/d196242e767fb9ccd076927b9c2f4f4dcf4ec153)

laserkelvin (approved):
Thanks for addressing my comments!

